Issue #8, opened 2017-02-12
one PQueue datatype with type parameter for priority order


amigalemming:
Currently you define `MaxPQueue` as wrapper around `MinPQueue`. I would prefer a general `PQueue` data type with a type parameter that chooses between the Min and Max variant.
That would save the wrapper code in `Data.PQueue.Max` and code duplication in custom utility functions.
I think it can still be Haskell 98 with definitions like these:
~~~~
class Dir dir where dirCompare :: (Ord a) => dir -> a -> a -> Ordering
data Min = Min; instance Dir Min where dirCompare Min = compare
data Max = Max; instance Dir Max where dirCompare Max = flip compare
~~~~

lspitzner:
I think I understand what you have in mind there and although I cannot really imagine the advantage of this yet, in general I currently see no issues with this approach either - performance impact of the additional indirection should be insignificant and it should be possible to implement it in a backwards-compatible fashion (right? function signatures would change but that change would be nothing more than a constraint on the type argument that can be hidden behind an alias..).
(To clarify: I only consider the haskell98 route here.)

If this is correct and existing users won't be disturbed by such a change, I'll happily merge a PR. However I won't implement this myself.

Sorry for the late answer; It took me a moment to realize how the haskell98-route would work and how exactly it would change the interface.

amigalemming:
On Thu, 2 Mar 2017, Lennart Spitzner wrote:

> 
> I think I understand what you have in mind there and although I cannot really imagine
> the advantage of this yet, in general I currently see no issues with this approach
> either - performance impact of the additional indirection should be insignificant and
> it should be possible to implement it in a backwards-compatible fashion (right?
> function signatures would change but that change would be nothing more than a
> constraint on the type argument that can be hidden behind an alias..).
> (To clarify: I only consider the haskell98 route here.)

It is not completely backwards compatible because type class instances on 
the current types would become instances on type synonyms. That is, users 
would need to add FlexibleInstances, or better, they would merge their 
instances of Max and Min queues into unified instances.

amigalemming:
On Thu, 2 Mar 2017, Lennart Spitzner wrote:

> If this is correct and existing users won't be disturbed by such a 
> change, I'll happily merge a PR. However I won't implement this myself.

I think now that I can do it in a completely backwards-compatible way. 
However, I am nervous that there is no test-suite that can check whether 
the modified code is still correct and some warnings are disabled. E.g. 
[...]

PR #14, opened 2017-03-12
New modules PQueue and PQueue.Prio where order type is a type parameter

thielema:
This is finally my proposed code for solving https://github.com/lspitzner/pqueue/issues/8.
I defined the type Wrap that is like the Down datatype
but it has a 'top' type parameter to choose
whether minimum or maximum should be at the top of the queue.
The new modules PQueue and PQueue.Prio are rather copies
of their Max counterparts, using Wrap instead of Down.
I like that approach very much! :-)
Thus I'd propose to phase out public PQueue.Min/Max and PQueue.Prio.Min/Max modules slowly,
that is, add new functions only to PQueue and PQueue.Prio,
and somewhen deprecate PQueue.Min/Max and PQueue.Prio.Min/Max,
and very far in the future remove Max modules and Down and make Min modules private.
It will save you code and Haddock duplication.
(As we have seen, it is pretty easy to get the duplicated Haddock comments wrong.)

### DIFF
--- a/Data/PQueue.hs
+++ b/Data/PQueue.hs
@@ -1,0 +1,394 @@
+{-# LANGUAGE CPP #-}
+
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Data.PQueue
+-- Copyright   :  (c) Henning Thielemann 2017
+--                (c) Louis Wasserman 2010
+-- License     :  BSD-style
+-- Maintainer  :  libraries@haskell.org
+-- Stability   :  experimental
+-- Portability :  portable
+--
+-- General purpose priority queue, supporting view-top operations.
+--
+-- An amortized running time is given for each operation, with /n/ referring
+-- to the length of the sequence and /k/ being the integral index used by
+-- some operations.  These bounds hold even in a persistent (shared) setting.
+--
+-- This implementation is based on a binomial heap augmented with a global root.
+-- The spine of the heap is maintained lazily.  To force the spine of the heap,
+-- use 'seqSpine'.
+--
+-- This implementation does not guarantee stable behavior.
+--
+-- This implementation offers a number of methods of the form @xxxU@, where @U@ stands for
+-- unordered.  No guarantees whatsoever are made on the execution or traversal order of
+-- these functions.
+-----------------------------------------------------------------------------
+module Data.PQueue (
+  Queue,
+  MinQueue,
+  MaxQueue,
+  -- * Basic operations
+  empty,
+  null,
+  size,
+  -- * Query operations
+  findTop,
+  getTop,
+  deleteTop,
+  deleteFindTop,
+  delete,
+  topView,
+  -- * Construction operations
+  singleton,
+  insert,
+  insertBehind,
+  union,
+  unions,
+  -- * Subsets
+  -- ** Extracting subsets
+  (!!),
+  take,
+  drop,
+  splitAt,
+  -- ** Predicates
+  takeWhile,
+  dropWhile,
+  span,
+  break,
+  -- * Filter/Map
+  filter,
+  partition,
+  mapMaybe,
+  mapEither,
+  -- * Fold\/Functor\/Traversable variations
+  map,
+  foldrAsc,
+  foldlAsc,
+  foldrDesc,
+  foldlDesc,
+  -- * List operations
+  toList,
+  toAscList,
+  toDescList,
+  fromList,
+  fromAscList,
+  fromDescList,
+  fromOrderedList,
+  -- * Unordered operations
+  mapU,
+  foldrU,
+  foldlU,
+  elemsU,
+  toListU,
+  -- * Miscellaneous operations
+  keysQueue,
+  seqSpine) where
+
+import qualified Data.PQueue.Min as Min
+import qualified Data.PQueue.Top as Top
+import qualified Data.PQueue.Prio as Prio
+import Data.PQueue.Top (Top, Wrap(Wrap, unwrap))
+
+import Control.DeepSeq (NFData(rnf))
+
+import Data.Functor ((<$>))
+import Data.Monoid (Monoid(mempty, mappend))
+import Data.Maybe (fromMaybe)
+import Data.Foldable (foldl, foldr)
+
+import Prelude hiding (null, foldr, foldl, take, drop, takeWhile, dropWhile, splitAt, span, break, (!!), filter)
+
+#ifdef __GLASGOW_HASKELL__
+import Text.Read (Lexeme(Ident), lexP, parens, prec,
+  readPrec, readListPrec, readListPrecDefault)
+import Data.Data
+#else
+build :: ((a -> [a] -> [a]) -> [a] -> [a]) -> [a]
+build f = f (:) []
+#endif
+
+-- | A priority queue with elements of type @a@.  Supports extracting the top element.
+-- Implemented as a wrapper around 'Min.MinQueue'.
+newtype Queue top a = Q (Min.MinQueue (Wrap top a))
+# if __GLASGOW_HASKELL__
+  deriving (Eq, Ord, Data, Typeable)
+# else
+  deriving (Eq, Ord)
+# endif
+
+type MinQueue = Queue Top.Min
+type MaxQueue = Queue Top.Max
+
+instance NFData a => NFData (Queue top a) where
+  rnf (Q q) = rnf q
+
+instance (Top top, Ord a, Show a) => Show (Queue top a) where
+  showsPrec p xs = showParen (p > 10) $
+    showString "fromOrderedList " . shows (toList xs)
+
+instance Read a => Read (Queue top a) where
+#ifdef __GLASGOW_HASKELL__
+  readPrec = parens $ prec 10 $ do
+    Ident "fromOrderedList" <- lexP
+    xs <- readPrec
+    return (fromOrderedList xs)
+
+  readListPrec = readListPrecDefault
+#else
+  readsPrec p = readParen (p > 10) $ \ r -> do
+    ("fromOrderedList",s) <- lex r
+    (xs,t) <- reads s
+    return (fromOrderedList xs,t)
+#endif
+
+instance (Top top, Ord a) => Monoid (Queue top a) where
+  mempty = empty
+  mappend = union
+
+-- | /O(1)/.  The empty priority queue.
+empty :: Queue top a
+empty = Q Min.empty
+
+-- | /O(1)/.  Is this the empty priority queue?
+null :: Queue top a -> Bool
+null (Q q) = Min.null q
+
+-- | /O(1)/.  The number of elements in the queue.
+size :: Queue top a -> Int
+size (Q q) = Min.size q
+
+-- | /O(1)/.  Returns the top element of the queue.  Throws an error on an empty queue.
+findTop :: Queue top a -> a
+findTop = fromMaybe (error "Error: findTop called on empty queue") . getTop
+
+-- | /O(1)/.  The top element of the queue, if there is one.
+getTop :: Queue top a -> Maybe a
+getTop (Q q) = unwrap <$> Min.getMin q
+
+-- | /O(log n)/.  Deletes the top element of the queue.  Does nothing on an empty queue.
+deleteTop :: (Top top, Ord a) => Queue top a -> Queue top a
+deleteTop (Q q) = Q (Min.deleteMin q)
+
+-- | /O(log n)/.  Extracts the top element of the queue.  Throws an error on an empty queue.
+deleteFindTop :: (Top top, Ord a) => Queue top a -> (a, Queue top a)
+deleteFindTop = fromMaybe (error "Error: deleteFindTop called on empty queue") . topView
+
+-- | /O(log n)/.  Extract the top element of the sequence, if there is one.
+topView :: (Top top, Ord a) => Queue top a -> Maybe (a, Queue top a)
+topView (Q q) = case Min.minView q of
+  Nothing -> Nothing
+  Just (Wrap x, q')
+          -> Just (x, Q q')
+
+-- | /O(log n)/.  Delete the top element of the sequence, if there is one.
+delete :: (Top top, Ord a) => Queue top a -> Maybe (Queue top a)
+delete = fmap snd . topView
+
+-- | /O(1)/.  Construct a priority queue with a single element.
+singleton :: a -> Queue top a
+singleton = Q . Min.singleton . Wrap
+
+-- | /O(1)/.  Insert an element into the priority queue.
+insert :: (Top top, Ord a) => a -> Queue top a -> Queue top a
+x `insert` Q q = Q (Wrap x `Min.insert` q)
+
+-- | Amortized /O(1)/, worst-case /O(log n)/.  Insert an element into the priority queue,
+--   putting it behind elements that compare equal to the inserted one.
+insertBehind :: (Top top, Ord a) => a -> Queue top a -> Queue top a
+x `insertBehind` Q q = Q (Wrap x `Min.insertBehind` q)
+
+-- | /O(log (min(n1,n2)))/.  Take the union of two priority queues.
+union :: (Top top, Ord a) => Queue top a -> Queue top a -> Queue top a
+Q q1 `union` Q q2 = Q (q1 `Min.union` q2)
+
+-- | Takes the union of a list of priority queues.  Equivalent to @'foldl' 'union' 'empty'@.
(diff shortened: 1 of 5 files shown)


lspitzner:
Ok I have considered this PR and its design. I have not fully understood all details but I can see how this proposal is a generalization of the existing Min/Max duplication and it matches what I had in mind after the discussion in the issue. However I have a number of reservations.

Most importantly, when I said

> If this is correct and existing users won't be disturbed by such a change, I'll happily merge a PR.

I meant that users should not be disturbed. No phase-outs, no deprecation.

With this constraint, this PR in its current form increases code duplication. On a related note, the Min/Max duplication primarily is duplication of the interface, and a generalization of the interface does not remove the complexity, but merely moves the burden to the consumer of the interface. In haskell there exist a good amount of general-purpose generalizations where this makes sense, but the generalization used here seems rather purpose-specific. As a consequence, I give the code-duplication argument a rather low subjective score, which conversely means that I am not completely opposed to adding this generic interface in addition to the existing Min/Max interfaces. I also see the possibility for turning the Max interface into a completely shallow wrapper around `Queue Max`, even when this is probably less of an improvement than what you had in mind.

For more formal matters, and this confused me somewhat: I thought your main purpose of (re)adding the test-suite was to test this code, yet you opened this PR without adapting the suite to cover the newly introduced code in any way. Am I missing something?

Regarding the new constructs (mainly the `Top` module):
- Some comments on the newly introduced constructs would be helpful. (Especially the type class - I generally dislike classes that lack any laws or explanations.)
- The names confused me somewhat: "Top" feels semantically close to "maximum", which makes it a bad choice for a thing that can me Max or Min. "First" would be a better choice imo. And "Wrap" is very close to the [`Tagged`](http://hackage.haskell.org/package/tagged-0.8.5/docs/Data-Tagged.html#t:Tagged) type - it differs only in the Ord-instance, as far as I can tell. I think this fact deserves a comment too, if not a rename.
- I am mildly certain that the newtypes (`Compare`, `ToList`, …) can be avoided. I won't discuss this until the more important issues are resolved.

thielema:
On Thu, 16 Mar 2017, Lennart Spitzner wrote:

> Most importantly, when I said
>
>       If this is correct and existing users won't be disturbed by such a change, I'll happily merge a PR.
> 
> I meant that users should not be disturbed. No phase-outs, no deprecation.

> I also see the possibility for turning the Max interface into a 
> completely shallow wrapper around Queue Max, even when this is probably 
> less of an improvement than what you had in mind.

I considered making Prio.Max a newtype wrapper around (PQueue Max), but 
that seemed to be more work and even more possibilities to make errors and 
you would still have duplication of Haddock comments. Then I thought, that 
the module might become superfluous anyway.

My original idea was to implement the generalization at the core. But that 
required invasive changes and more complicated types. Thus I stopped that 
attempt.

> As a consequence, I give the code-duplication argument a rather low 
> subjective score, which conversely means that I am not completely 
> opposed to adding this generic interface in addition to the existing 
> Min/Max interfaces.

My proposed code does not save code at the library side, and it cannot, 
given that the existing interface shall be preserved and that it would 
require at least a substantial amount of wrapper code. However, it saves 
code on the client side. Every utility function in client code can now be 
defined uniformly for Max or Min.

If you want to avoid code duplication in the library, too, we could use a 
generic wrapper module that we duplicate using CPP or some semi-automatic 
process. We could add this in a further step.


> For more formal matters, and this confused me somewhat: I thought your 
> main purpose of (re)adding the test-suite was to test this code, yet you 
> opened this PR without adapting the suite to cover the newly introduced 
> code in any way.

It's not complete, yes. My confidence so far is entirely based on the fact 
that I did only search&replace on the Max modules.


> Regarding the new constructs (mainly the Top module):
>
>  *  Some comments on the newly introduced constructs would be helpful. (Especially the type class - I generally
>     dislike classes that lack any laws or explanations.)

Ok. The only law is that the class can only have Min or Max as instances. 
Users cannot add more instances (unless they use "undefined"). There are 
no more laws. That's what I call a closed-world class.

>  *  The names confused me somewhat: "Top" feels semantically close to "maximum", which makes it a bad choice for
>     a thing that can me Max or Min. "First" would be a better choice imo. And "Wrap" is very close to the Tagged
>     type - it differs only in the Ord-instance, as far as I can tell. I think this fact deserves a comment too,
>     if not a rename.

It depends on whether you imagine vertical or horizontal queues. I am 
happy with a horizontal queue, too. Unfortunately Git does not support 
renaming identifiers like Darcs does.

>  *  I am mildly certain that the newtypes (Compare, ToList, …) can be avoided. I won't discuss this until the
>     more important issues are resolved.

These newtypes are required by the closed-world approach. The idea of the 
closed-world class is that you cannot add more instances, but in return 
you can add any number of "methods" using these helper newtypes. The class 
API does not change by adding such "methods" (interesting for PVP 
compliance) and even a client can add "methods" without the need to define 
sub-classes.

lspitzner:
> My proposed code does not save code at the library side, and it cannot, given that the existing interface shall be preserved and that it would require at least a substantial amount of wrapper code. 

But you implied that it did when you said

> It will save you code and Haddock duplication.
(As we have seen, it is pretty easy to get the duplicated Haddock comments wrong.)

thielema:
On Thu, 16 Mar 2017, Lennart Spitzner wrote:

>       My proposed code does not save code at the library side, and it cannot, given that the existing
>       interface shall be preserved and that it would require at least a substantial amount of wrapper
>       code.
> 
> But you implied that it did when you said
>
>       It will save you code and Haddock duplication.
>       (As we have seen, it is pretty easy to get the duplicated Haddock comments wrong.)


I wrote that with phasing out Min/Max modules in mind. If you do not want 
that way, we will necessarily have code duplication.

lspitzner:
Ok. I guess I merely find it confusing that you make a (apparently new) code-duplication argument in response to what was said before.

lspitzner:
> It depends on whether you imagine vertical or horizontal queues. I am happy with a horizontal queue, too. Unfortunately Git does not support renaming identifiers like Darcs does.

I don't see how "first" is horizontal (?) And is Darcs vs Git relevant at all for this discussion?

thielema:
On Thu, 16 Mar 2017, Lennart Spitzner wrote:

>       It depends on whether you imagine vertical or horizontal queues. I am happy with a horizontal queue,
>       too. Unfortunately Git does not support renaming identifiers like Darcs does.
> 
> I don't see how "first" is horizontal (?)

You are right. That would have been "left".

> And is Darcs vs Git relevant at all for this discussion?

In Darcs it would be simpler to apply your proposed renaming afterwards. 
:-)

lspitzner:
Which remains irrelevant. Sorry.

lspitzner:
Ok, back to more constructive feedback: Let me explain what I have in mind to avoid the newtypes, one sec.

thielema:
On Thu, 16 Mar 2017, Lennart Spitzner wrote:

> Ok. I guess I merely find it confusing that you make a (apparently new) 
> code-duplication argument in response to what was said before.

When I opened the ticket, I hoped for reduction of code duplication 
because I expected that we can have one core generic module and make Min 
and Max thin wrappers around it. Then I found there is no test-suite that 
could check my modifications. Then I found a test-suite, but it lacked to 
check all modules and all functions. But I added it, because it is better 
than nothing. Then I found that it is not at all easy to generalize the 
core implementation and decided to keep the chain Internals->Min->Max, but 
copy Max to Top. Now, code un-duplication can only occur, if we phase out 
Max. Now, since this is not wanted, too, we can only have code savings on 
the client side.

lspitzner:
I understand, and I can see that the whole process may be frustrating for you. However I still think that I have properly communicated the constraint that I would put on this PR early on.

I'll say this again to make it clear: I am not opposed to merging this on the basis that it increases duplication; I was actually arguing _for_ duplication in my initial reply to this PR.

Am I assuming correctly that you still want this merged, even when we do not deprecate the existing interface?

thielema:
On Thu, 16 Mar 2017, Lennart Spitzner wrote:

> I understand, and I can see that the whole process may be frustrating 
> for you. However I still think that I have properly communicated the 
> constraint that I would put on this PR early on.

You said:
    "If this is correct and existing users won't be disturbed by such a change,"

I assumed that by no longer adding public functions to Min/Max you would 
not disturb existing users. You can make the other phases of deprecation 
as long as you want. :-)

> Am I assuming correctly that you still want this merged, even when we do 
> not deprecate the existing interface?

Right.

lspitzner:
>  I assumed that by no longer adding public functions to Min/Max you would not disturb existing users. You can make the other phases of deprecation as long as you want. :-)

Ok, fair :-) Then we are on the same page.

lspitzner:
When defining ~~Top~~ First like this:

~~~~.hs
class First first where
  switch :: a -> a -> TaggedF first a

instance First Min where switch f _ = TaggedF f
instance First Max where switch _ f = TaggedF f
~~~~

we can avoid the `Compare` newtype like this:

~~~~.hs
instance (First first, Ord a) => Ord (TaggedF first a) where
  compare x y = unTaggedF $ switch compare (flip compare) <*> x <*> y
~~~~

And by defining

~~~~.hs
unSwitchQueue
  :: (Min.MinQueue (TaggedF first a) -> TaggedF first b) -> Queue first a -> b
unSwitchQueue f (Q q) = unTaggedF (f q)
~~~~

we can get rid of `Foldr`, `Foldl`, `ToList` and write, e.g.
~~~~.hs
foldrDesc f z = unSwitchQueue $ \q ->
  switch (Min.foldrDesc (f . unTaggedF) z q) (Min.foldrAsc (f . unTaggedF) z q)
~~~~

I haven't tested this either, but it should have the same semantics; it at least compiles so far. Or is there some inherent disadvantage to this approach which I miss? Using Applicative or appropriate general-purpose helpers (unswitch) seem to serve the purpose of "connecting the phantom types" just as well as the newtypes.

lspitzner:
(`unswitch` may very well be a bad name; it acts together with switch in these cases but perhaps implying that it is some kind of dual is wrong.)

lspitzner:
I would have committed the changes I made already, but they don't compile yet. I have pushed the not-yet-compiling code to my [generic-top](https://github.com/lspitzner/pqueue/tree/generic-top) branch.

thielema:
On Thu, 16 Mar 2017, Lennart Spitzner wrote:

> When defining Top First like this:
> 
> class First first where
>   switch :: a -> a -> TaggedF first a
> 
> instance First Min where switch f _ = TaggedF f
> instance First Max where switch _ f = TaggedF f
> 
> we can avoid the Compare newtype like this:
> 
> instance (First first, Ord a) => Ord (TaggedF first a) where
>   compare x y = unTaggedF $ switch compare (flip compare) <*> x <*> y
> 
> And by defining
> 
> unSwitchQueue
>   :: (Min.MinQueue (TaggedF first a) -> TaggedF first b) -> Queue first a -> b
> unSwitchQueue f (Q q) = unTaggedF (f q)
> 
> we can get rid of Foldr, Foldl, ToList and write, e.g.

Does your approach also work for functions with TaggedF as result, e.g. 
FromList? Maybe it works and you need an extra unSwitchQueue instead of an 
extra newtype for every new type signature?

That said, we could get rid of Foldr, Foldl and ToList by using a unified 
newtype wrapper, say,

newtype Fold b a first = Fold {runFold :: Queue first a -> b}

and adapt foldr, foldl and toList to match this interface. We would save 
two newtype definitions and get in return a bit of adaption code that must 
be duplicated for the Min and the Max branch in 'switch'. You already 
write this duplicated adaption code (here: unTaggedF) in your example to 
match the generic unSwitchQueue:

> foldrDesc f z = unSwitchQueue $ \q ->
>   switch (Min.foldrDesc (f . unTaggedF) z q) (Min.foldrAsc (f . unTaggedF) z q)

thielema:
On Thu, 16 Mar 2017, Lennart Spitzner wrote:

> I would have committed the changes I made already, but they don't 
> compile yet. I have pushed the not-yet-compiling code to my generic-top 
> branch.

I think that your approach and my one so far are pretty equivalent. I can 
convert your 'compare' implementation and your 'unSwitchQueueConstr' to my 
newtype style and you can express with functions what I express with 
types. I use the newtypes to select the type to switch on, and you use 
functions for that. I have to think a bit more about it, maybe your 
approach using functions is more flexible. That would be cool and would be 
useful for other packages where I use closed-world classes, e.g. pathtype.

lspitzner:
True, this still requires 2 helper functions in place of the 5 newtypes (I don't count unTaggedF as a helper function; the newtype-approach requires `runFoo` after all in addition to the newtype definition). I still like the newtype-free approach because there are only two "carriers" for the phantom type: TaggedF and Q, and the helper functions have the clear purpose of forcing the phantom types of these two types to line up.

lspitzner:
Btw, can we merge the Private and Internals modules?

thielema:
On Thu, 16 Mar 2017, Lennart Spitzner wrote:

> True, this still requires 2 helper functions in place of the 5 newtypes

As I said, I can convert your implementations to the newtype style, and 
then there remain only 2 newtypes, one corresponding to unSwitchQueue and 
one corresponding to unSwitchQueueConstr.

thielema:
On Thu, 16 Mar 2017, Lennart Spitzner wrote:

> Btw, can we merge the Private and Internals modules?

Prio.Internals is the private module for Prio.Min and Prio.Private is the 
private module for Prio. They could have better names, but I would not 
merge them.

lspitzner:
> As I said, I can convert your implementations to the newtype style, and then there remain only 2 newtypes, one corresponding to unSwitchQueue and one corresponding to unSwitchQueueConstr.

Right. the difference is indeed fairly small.

Both this suggestion and the question of how to name things don't hold back this PR. I'd like proper testing and some comments (and I was just as guilty when not adding my reasoning for `TaggedF` vs `Tagged` in the code). The newtypes are not even user-facing, so we'd be able to switch at any point anyways. Only "top" vs "first" is user-facing, thinking of it.

lspitzner:
> Prio.Internals is the private module for Prio.Min and Prio.Private is the private module for Prio. They could have better names, but I would not merge them.

Good point. Also deserves a comment at the top of these two modules :-)

thielema:
On Thu, 16 Mar 2017, Lennart Spitzner wrote:

>       Prio.Internals is the private module for Prio.Min and Prio.Private is the private module for Prio.
>       They could have better names, but I would not merge them.
> 
> Good point. Also deserves a comment at the top of these two modules :-)

I'd prefer better module names but hesitated to rename them so far, 
because that would be more invasive and erm, Git does not directly support 
moving files, in contrast to Darcs. :-)

lspitzner:
haha :-)

thielema:
On Thu, 16 Mar 2017, Lennart Spitzner wrote:

>       As I said, I can convert your implementations to the newtype style, and then there remain only 2
>       newtypes, one corresponding to unSwitchQueue and one corresponding to unSwitchQueueConstr.
> 
> Right. the difference is indeed fairly small.

There is one small difference: My approach asserts that the two instances 
of Top are correct and not swapped.

I tried to prove that both approaches are equivalent. I assume they are 
equivalent in our case, where Min and Max are phantoms. But I am afraid 
that I cannot generalize it to my other projects.

Here is my attempt to prove equivalence:

~~~~
module ClosedWorld where

data Min = Min
data Max = Max


class Top top where
    switchTop :: f Min -> f Max -> f top

instance Top Min where switchTop f _ = f
instance Top Max where switchTop _ f = f


switchTopFromFirst :: (First top) => f Min -> f Max -> f top
switchTopFromFirst = undefined switchFirst



newtype Tagged first a = Tagged {untag :: a}

class First first where
    switchFirst :: a -> a -> Tagged first a

instance First Min where switchFirst a _ = Tagged a
instance First Max where switchFirst _ a = Tagged a


newtype Const a first = Const a

taggedFromConst :: Const a first -> Tagged first a
taggedFromConst (Const x) = Tagged x

switchFirstFromTop :: (Top first) => a -> a -> Tagged first a
switchFirstFromTop x y = taggedFromConst $ switchTop (Const x) (Const y)
~~~~

We would need to implement switchTopFromFirst with the given type 
signature using only switchFirst.

lspitzner:
I agree, in general `switch :: f Min -> f Max -> f top` can provide stronger guarantees. But when using freshly constructed newtypes, you get no additional safety (if you switch the arguments to `switch` in this PR, it still compiles in a few cases that I have tested at least).

Now I wonder if one can avoid newtypes even with that.. I have to play with this a bit. (And no, newtypes are not evil and this is probably not very important.)

treeowl:
I've been skimming comments, but ... I'm so far failing to understand the point of any of this. What's wrong with just turning a concrete min-priority queue into a max-priority queue using `Down`? It seems dead simple already.

PR #15, opened 2017-07-31
Bumping base upper bounds to support ghc 8.2

mstksg:
Package seems to compile fine with base-4.10 :)  Just bumping the upper bound to 4.11 so it can be built with ghc 8.2, if you wanted!

Thanks for the great package!


lspitzner:
Thanks! Will release on hackage later today.

PR #44, opened 2021-12-06
Add GitHub CI

konsumlamm:
I added GHC versions 7.10.3, 8.0.2, 8.2.2 and 8.4.4 to the `tested-with` section (since we have `base >= 4.8`, so the minimum GHC version is 7.10). The CI tests all versions from the `tested-with` section.

BTW, I suggest merging PRs using "squash and merge", which combines all PR commits into a single one and commits that (without an annoying "Merge pull request" commit).

### DIFF
--- a/.github/workflows/haskell.yaml
+++ b/.github/workflows/haskell.yaml
@@ -1,0 +1,21 @@
+name: Haskell CI
+
+on:
+  - push
+  - pull_request
+
+jobs:
+  build:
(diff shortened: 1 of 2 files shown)


treeowl:
There's also "rebase and merge", which keeps separate commits and just sticks them on.

treeowl:
I definitely agree we should test everything we support, as long as we're not scared to drop older GHC versions when that's helpful.

PR #45, opened 2021-12-07
Improve RULES situation

### LINKED ISSUE #4: -Wall cleanliness: RULE pragma warnings
I disable this warning in the .cabal as it is probably harmless. Should be fixed properly at some point.

treeowl:
Add staging to make rules more likely to work out properly
and get rid of rule/inline warnings.

Fixes #4

### DIFF
--- a/src/Data/PQueue/Min.hs
+++ b/src/Data/PQueue/Min.hs
@@ -232,15 +232,25 @@ partition p = mapEither (\x -> if p x then Left x else Right x)
 map :: Ord b => (a -> b) -> MinQueue a -> MinQueue b
 map f = foldrU (insert . f) empty
 
-{-# INLINE toAscList #-}
+{-# INLINABLE [1] toAscList #-}
 -- | /O(n log n)/. Extracts the elements of the priority queue in ascending order.
 toAscList :: Ord a => MinQueue a -> [a]
-toAscList queue = build (\c nil -> foldrAsc c nil queue)
+toAscList queue = foldrAsc (:) [] queue
 
-{-# INLINE toDescList #-}
+{-# INLINABLE toAscListApp #-}
+toAscListApp :: Ord a => MinQueue a -> [a] -> [a]
+toAscListApp Empty app = app
+toAscListApp (MinQueue _ x ts) app = x : foldrUnfold (:) app extractHeap ts
+
+{-# INLINABLE [1] toDescList #-}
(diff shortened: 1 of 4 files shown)


src/Data/PQueue/Min.hs
```diff
@@ -248,17 +259,19 @@ toList :: Ord a => MinQueue a -> [a]
 toList = toAscList
 
 {-# RULES
-  "toAscList" forall q . toAscList q = build (\c nil -> foldrAsc c nil q);
-    -- inlining doesn't seem to be working out =/
-  "toDescList" forall q . toDescList q = build (\c nil -> foldrDesc c nil q);
-  #-}
+"toAscList" [~1] forall q . toAscList q = build (\c nil -> foldrAsc c nil q)
+"toDescList" [~1] forall q . toDescList q = build (\c nil -> foldrDesc c nil q)
```
konsumlamm, line 263:
```suggestion
"toAscList" [~1] forall q. toAscList q = build (\c nil -> foldrAsc c nil q)
"toDescList" [~1] forall q. toDescList q = build (\c nil -> foldrDesc c nil q)
```

konsumlamm, line 263:
```suggestion
"toAscList" [~1] forall q. toAscList q = build (\c nil -> foldrAsc c nil q)
"toDescList" [~1] forall q. toDescList q = build (\c nil -> foldrDesc c nil q)
```

konsumlamm (changes requested):
I've never used staging or rewrite rules, so I can't comment on that. I don't really like rewrite rules, but this just modifies existing ones, so I have no problem.

I see you've added `Data.PQueue.Min.foldlAsc'`, that should be added to the changelog. Is there any reason you didn't add `foldrAsc'`, `foldlDesc'` etc? I don't see any obvious reason why `foldlAsc'` should be special.

treeowl:
> I see you've added `Data.PQueue.Min.foldlAsc'`, that should be added to the changelog. Is there any reason you didn't add `foldrAsc'`, `foldlDesc'` etc? I don't see any obvious reason why `foldlAsc'` should be special.

For min-queues, the natural/efficient order-respecting folds are `foldrAsc`, `foldlAsc'`, `foldlDesc`, and `foldrDesc'`. I'm not convinced we want the wonky ones we already have (`foldlAsc` and `foldrDesc`) let alone any new ones. But you're right; I should deal with that all in a separate PR.

PR #46, opened 2021-12-07
Foldable traversable

treeowl:
Improve folding and traversing. Add `mapMWithKey` (for traversing in strict monads). Clean up cruft.


src/Data/PQueue/Prio/Max.hs
```diff
@@ -298,9 +305,24 @@ foldlWithKey f z0 (MaxPQ q) = Q.foldlWithKey (\z -> f z . unDown) z0 q
 -- (@'traverseWithKey' f q == 'fromDescList' <$> 'traverse' ('uncurry' f) ('toDescList' q)@)
 --
 -- If you do not care about the /order/ of the traversal, consider using 'traverseWithKeyU'.
+--
+-- If you are working in a strict monad, consider using 'mapMWithKey'.
 traverseWithKey :: (Ord k, Applicative f) => (k -> a -> f b) -> MaxPQueue k a -> f (MaxPQueue k b)
 traverseWithKey f (MaxPQ q) = MaxPQ <$> Q.traverseWithKey (f . unDown) q
 
+-- | A version of 'traverseWithKey' designed to be efficient for strict monads
+-- such as 'IO' and strict @State@.
```
konsumlamm, line 314: What exactly qualifies a strict monad? That should be documented.

src/Data/PQueue/Prio/Max.hs
```diff
@@ -298,9 +305,24 @@ foldlWithKey f z0 (MaxPQ q) = Q.foldlWithKey (\z -> f z . unDown) z0 q
...
+mapMWithKey :: (Ord k, Monad m) => (k -> a -> m b) -> MaxPQueue k a -> m (MaxPQueue k b)
+mapMWithKey f = go empty
+  where
+    go acc q = acc `seq`
```
konsumlamm, line 318:
```suggestion
    go !acc q =
```
If we already enable `BangPatterns`, we might as well use them.

src/Data/PQueue/Prio/Max.hs
```diff
@@ -298,9 +305,24 @@ foldlWithKey f z0 (MaxPQ q) = Q.foldlWithKey (\z -> f z . unDown) z0 q
...
+      case maxViewWithKey q of
+        Nothing           -> pure acc
+        Just ((k, a), q') -> do
+          b <- f k a
+          let !acc' = insertMin' k b acc
```
konsumlamm, line 323:
```suggestion
          let !acc' = insertMax' k b acc
```
`k` is the maximum key, so this should use `insertMax'`, no?

src/Data/PQueue/Prio/Max/Internals.hs
```diff
@@ -45,3 +47,6 @@ instance Foldable Down where
 
 instance Traversable Down where
   traverse f (Down a) = Down <$> f a
+
+insertMin' :: k -> a -> MaxPQueue k a -> MaxPQueue k a
+insertMin' k a (MaxPQ q) = MaxPQ (Internals.insertMax' (Down k) a q)
```
konsumlamm, line 52: I don't see why this is needed, see above.

konsumlamm (commented):
I'm not sure I'm convinced about the need for `mapMWithKey`. I don't know of any similar function in other related packages. Do you have benchmarks that show a significant speedup?

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Revision history for pqueue
 
+## 1.4.2
+
+  * Large performance improvements, especially when there are many insertions
+    in a row. Insertion, deletion, and merge are now *worst case* logarithmic,
+    while maintaining their previous amortized bounds.
```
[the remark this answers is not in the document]
treeowl: It is. That's a good idea.

pqueue.cabal
```diff
@@ -42,6 +41,10 @@ library
   if impl(ghc) {
     default-extensions: DeriveDataTypeable
   }
+  other-extensions:
+      BangPatterns
+    , CPP
+    , StandaloneDeriving
```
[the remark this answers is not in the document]
treeowl: I think it's reasonable not to commit utterly to GHC. The extensions we use unconditionally (so far) will most likely be available in any Report-style Haskell implementation (PureScript strays too far to support, unfortunately).

src/Data/PQueue/Internals.hs
```diff
@@ -46,12 +46,6 @@ import Prelude hiding (null)
 
 -- | A priority queue with elements of type @a@. Supports extracting the minimum element.
 data MinQueue a = Empty | MinQueue {-# UNPACK #-} !Int !a !(BinomHeap a)
-#if __GLASGOW_HASKELL__>=707
-  deriving Typeable
-#else
-#include "Typeable.h"
-INSTANCE_TYPEABLE1(MinQueue,minQTC,"MinQueue")
-#endif
```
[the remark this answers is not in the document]
treeowl: It's derived automatically precisely from 7.10 onward.

src/Data/PQueue/Prio/Max.hs
treeowl: A strict monad is one for which `_|_ >>= pure () = _|_`. I'm not sure that *precisely* characterizes where it makes sense to do this, TBH, but I think it's a decent heuristic. Have you looked at the code to understand what it's about?

src/Data/PQueue/Prio/Max.hs
treeowl: Yes, true. There are a *couple* spots in the code where I'd prefer to keep the `seq` explicit (forcing on cascade), but this isn't one of them.

src/Data/PQueue/Prio/Max.hs
treeowl: It's the maximum of (what remains of) the original queue. It's the minimum of the accumulator.

src/Data/PQueue/Prio/Max/Internals.hs
treeowl: It is; see above.

src/Data/PQueue/Prio/Max.hs
treeowl: That link isn't working for me. What does the comment say?

konsumlamm: It's the one about `insertMin'`, you already replied to it.

src/Data/PQueue/Prio/Max.hs
konsumlamm: I see, you're right.

treeowl:
@konsumlamm, here's a test program:

```haskell
{-# language BangPatterns #-}
module Main where
import Data.PQueue.Prio.Min (MinPQueue)
import qualified Data.PQueue.Prio.Min as Q
import Control.Monad.Trans.State.Strict

glump :: MinPQueue Int Int
glump = Q.fromList [(a, a) | a <- [1..2*10^6]]

hoom :: MinPQueue Int Int -> MinPQueue Int Int
hoom = flip evalState 0 . Q.mapMWithKey
  (\x y -> do
     !old <- get
     put (old + 1)
     pure (x + y + old))

main = print $ take 3 $ Q.toAscList (hoom glump)
```

On my system, this runs in (give or take)

```
real    0m1.688s
user    0m1.519s
sys     0m0.168s
```

If I use `traverseWithKey` instead, it runs in (give or take)

```
real    0m2.620s
user    0m2.363s
sys     0m0.256s
```

That seems to me like enough to care about.

src/Data/PQueue/Prio/Max.hs
```diff
@@ -172,8 +176,15 @@ instance Ord k => Foldable (MaxPQueue k) where
...
 
+  length = size
+  null = null
+
+-- | Traverses in descending order. 'mapM' is strictly accumulating like
+-- 'mapMWithKey'.
 instance Ord k => Traversable (MaxPQueue k) where
   traverse f (MaxPQ q) = MaxPQ <$> traverse f q
+  mapM = mapMWithKey . const
+  sequence = mapM id
```
konsumlamm, line 187: Is that really lawful for non-strict monads?

konsumlamm (commented):
Next time, please wait until the PR has been approved/re-reviewed before you merge, to make sure that the comments really have been addressed.

src/Data/PQueue/Prio/Max.hs
treeowl: Non-strict monads aren't lawful themselves. Consider lazy `State`.

```haskell
newtype State s a = State {runState :: s -> (a, s)}
instance Functor (State s) where
  -- This is equivalent to liftM
  fmap f (State st) = State $ \s ->
    case st s of
      ~(a, s') -> (f a, s')
```

This violates the functor identity law: `fmap id m = m`.

```haskell
runStateT (fmap id undefined) ()
=
case undefined () of
  ~(a, s') -> (id a, s')
=
(undefined, undefined)
```

But

```haskell
runStateT undefined () = undefined
```

There's always a certain amount of wishy-washiness when dealing with non-strict monads. I think it's perfectly fine to lean on "fast and loose reasoning is morally correct" and go with it. Note also that when you "run" a non-strict monad, you're almost always doing so *because you want a result*. But `traverseWithKey` won't give you a result when `mapMWithKey` doesn't.

PR #47, opened 2021-12-07
Fix warnings & implement missing method

konsumlamm:
Now [-Wnoncanonical-monoid-instances](https://downloads.haskell.org/ghc/latest/docs/html/users_guide/using-warnings.html#ghc-flag--Wnoncanonical-monoid-instances) doesn't complain about `mappend` not being defined as `mappend = (<>)` anymore. Also added missing `mconcat = unions` for `Data.PQueue.Max`.


treeowl (approved):
I greatly dislike `mconcat` as a class method, but yes, we should do this.

Issue #50, opened 2021-12-07
Data instances are unsafe/wrong


treeowl:
It strikes me as somewhat peculiar that we have `Data` instances for priority queues at all, but if we're to do so, they should be consistent!

```haskell
{-# language ScopedTypeVariables, GADTs, TypeOperators, TypeApplications #-}

import Data.Data
import Data.Typeable
import Data.PQueue.Min

morp :: MinQueue Int -> MinQueue Int
morp = gmapT go
  where
    go :: forall d. Data d => d -> d
    go d
      | Just Refl <- eqT @d @Int
      = abs d
      | Just Refl <- eqT @d @(MinQueue Int)
      = d

main = print $ morp $ fromList [-5 .. 5]
```

This prints

```
fromAscList [5,-4,-3,-2,-1,0,1,2,3,4,5]
```

What's happening here? The `Data` instance for `MinQueue` (roughly speaking) pretends that we have

```haskell
data MinQueue a = empty | a <| MinQueue a
```

(Yes, those "constructor" names are lower case on purpose; that's basically okay.) The user can edit the "minimum value" and "the rest of the queue" independently, and the `Data` instance just assumes that the new minimum is in fact now greater than the minimal element of the new rest of the queue. That's ... peculiar.

----

Now we get to the inconsistency. `Data.PQueue.Prio.Min` has a similarly peculiar but totally *different* view of the structure of a priority queue! Specifically, it pretends we have (roughly)

```haskell
data MinpQueue a b = MinPQueue [(a, b)]
```

(but blows up if you try to *construct* a `MinPQueue` using `gunfold` or to inspect one using `toConstr`). Bzzwhat?

----

`MaxQueue` and `MaxPQueue` each use *derived* `Data` instances. That means they inherit all the peculiarities of the underlying instances, as well as revealing the relationships between the queue types. But wait! There's more! The `Down` type used to implement those types *isn't exported from the package*. This makes anything `syb`-style on it a bit ... awkward.

### Solutions

The easiest solution is to delete all the `Data` instances. I know, I know, that could break things. Other things we could do:

- 1\. Structure the `MinPQueue` instance more like the `MinQueue` instance. I think the latter is much nicer (because it doesn't force the queue to be completely rebuilt for no reason).
- 2a\. Make the instances preserve the queue invariants.
- 2b\. Document that the instances don't preserve the queue invariants.
[...]

PR #51, opened 2021-12-07
Fix Data instances

treeowl:
* Make the `Data` instances preserve the order invariant.

* Implement the rest of the methods for the `MinPQueue` instance.

This begins the work on #50.

### DIFF
--- a/src/Data/PQueue/Internals.hs
+++ b/src/Data/PQueue/Internals.hs
@@ -54,7 +54,7 @@ data MinQueue a = Empty | MinQueue {-# UNPACK #-} !Int !a !(BinomHeap a)
 instance (Ord a, Data a) => Data (MinQueue a) where
   gfoldl f z q = case minView q of
     Nothing      -> z Empty
-    Just (x, q') -> z insertMinQ `f` x `f` q'
+    Just (x, q') -> z insert `f` x `f` q'
 
   gunfold k z c = case constrIndex c of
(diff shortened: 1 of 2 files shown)


src/Data/PQueue/Internals.hs
```diff
@@ -70,7 +70,7 @@ instance (Ord a, Data a) => Data (MinQueue a) where
   dataTypeOf _ = queueDataType
 
 queueDataType :: DataType
-queueDataType = mkDataType "Data.PQueue.Min.MinQueue" [emptyConstr, consConstr]
+queueDataType = mkDataType "Data.PQueue.Internals.MinQueue" [emptyConstr, consConstr]
```
konsumlamm, line 73: I'm not sure about this. Of course it is actually defined in `Data.PQueue.Internals`, but logically it's defined in `Data.PQueue.Min`. Users shouldn't need to care about the internal modules imo.

treeowl: I got nervous about [`liftData`](https://hackage.haskell.org/package/template-haskell-2.18.0.0/docs/Language-Haskell-TH-Syntax.html#v:liftData), but fortunately that gets its module (and package) names from the `TypeRep`, not from any `Data` junk. So it's probably fine to leave that as `Data.PQueue.Min`; I'll make the change.

konsumlamm (approved):

PR #53, opened 2021-12-07
Remove all orphan instances

### LINKED ISSUE #52: Remove orphan instances
I really don't like orphan instances. Let's get rid of them.

treeowl:
* Remove all orphan instances.

* One of the `unions` definitions used `foldl`; make it use `foldl'`
  instead.

Closes #52.


treeowl:
@konsumlamm This PR moves a *lot* of things around, so I'm pretty much blocked on everything else till it goes in.

src/Data/PQueue/Prio/Max.hs
```diff
@@ -116,394 +112,5 @@ module Data.PQueue.Prio.Max (
...
--- | /O(log n)/. @seqSpine q r@ forces the spine of @q@ and returns @r@.
---
--- Note: The spine of a 'MaxPQueue' is stored somewhat lazily. Most operations
--- take great care to prevent chains of thunks from accumulating along the
--- spine to the detriment of performance. However, 'mapKeysMonotonic' can leave
--- expensive thunks in the structure and repeated applications of that function
--- can create thunk chains.
-seqSpine :: MaxPQueue k a -> b -> b
-seqSpine (MaxPQ q) = Q.seqSpine q
+import Prelude ()
```
konsumlamm, line 116: What is this good for?

src/Data/PQueue/Prio/Max.hs
```diff
@@ -116,394 +112,5 @@ module Data.PQueue.Prio.Max (
...
-empty = MaxPQ Q.empty
-
--- | /O(1)/. Constructs a singleton priority queue.
-singleton :: k -> a -> MaxPQueue k a
-singleton k a = MaxPQ (Q.singleton (Down k) a)
-
--- | Amortized /O(1)/, worst-case /O(log n)/. Inserts
--- an element with the specified key into the queue.
-insert :: Ord k => k -> a -> MaxPQueue k a -> MaxPQueue k a
-insert k a (MaxPQ q) = MaxPQ (Q.insert (Down k) a q)
```
konsumlamm, line 202: Why did you move all the functions to `Data.PQueue.Prio.Max` instead of only the instances (and fucntions needed for the instances)?

konsumlamm (commented):
Are the `Functor`, `Fodlable` and `Traversable` instances of `Down` used anywhere? Also, I think a better module name instead of `Data.PQueue.Prio.Max.Down` would be `Data.PQueue.Down`, since it's not specific to `MaxQueue` or `MaxPQueue`.

src/Data/PQueue/Prio/Max.hs
treeowl: It was much, much easier.

src/Data/PQueue/Prio/Max.hs
treeowl: I doubt it's good for much. It was here before #26, which made it *mostly* redundant. `mapKeysMonotonic` suspends O(n) work. Under certain circumstances, you might want to force that work to happen immediately or at a particular time.

treeowl: Oh, misread your comment. We don't want the `Prelude`, which exports various names that clash with ours.

treeowl:
> Are the `Functor`, `Fodlable` and `Traversable` instances of `Down` used anywhere?

I don't know.

> Also, I think a better module name instead of `Data.PQueue.Prio.Max.Down` would be `Data.PQueue.Down`, since it's not specific to `MaxQueue` or `MaxPQueue`.

Good point.

src/Data/PQueue/Prio/Max/Internals.hs
```diff
@@ -1,23 +1,146 @@
...
+-- We do not guarantee stable behavior.
+-- Ties are broken arbitrarily -- that is, if @k1 <= k2@ and @k2 <= k1@, then there
+-- are no guarantees about the relative order in which @k1@, @k2@, and their associated
+-- elements are returned. (Unlike Data.Map, we allow multiple elements with the
+-- same key.)
+--
+-- This implementation offers a number of methods of the form @xxxU@, where @U@ stands for
+-- unordered. No guarantees whatsoever are made on the execution or traversal order of
+-- these functions.
+-----------------------------------------------------------------------------
```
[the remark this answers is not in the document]
treeowl: Uhhh...... Maybe I didn't need to do that.

src/Data/PQueue/Prio/Max.hs
konsumlamm: Oh, right, didn't think there.

src/Data/PQueue/Prio/Max/Internals.hs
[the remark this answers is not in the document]
treeowl: Removed.

treeowl:
`Traversable Down` wasn't used, so I removed it. `Functor Down` and `Foldable Down` are currently used. We probably don't actually want to use `Foldable Down` (see #55), but we'll deal with that in a different PR. The `Functor` instance is used by things like `mapKeys` and `mapKeysMonotonic`. These are pretty inefficient no matter what we do (short of making all the trees lazy, which I'd very much rather not do), and I'm not optimistic that we can improve them much by reducing dictionary costs, but I guess we could try. I renamed the `Down` module `Data.PQueue.Internal.Down`, since it really is an internal thing.

konsumlamm (approved):

PR #58, opened 2021-12-09
Enhance parallel building

treeowl:
Move the `keysQueue` implementation out of `Data.PQueue.Internals`.
This allows that module to build in parallel with
`Data.PQueue.Prio.Internals`. It looks like [Cabal can't yet make use of
this](https://github.com/haskell/cabal/issues/976), but work to make
it do so is under way.


konsumlamm (approved):
Seems a bit pointless when parallel building isn't supported yet, but sure, why not.

Issue #55, opened 2021-12-07
Make sure we're not building unnecessarily large dictionaries


treeowl:
Things like traversals and (especially) folds can easily get out of hand with nonregular types. We should make sure they don't.

konsumlamm:
Can you elaborate a bit? What dictionaries exactly are you talking about and why could they get unnecessarily large?

treeowl:
I need to review the code to see if there's actually an issue or not. Instances for non-regular types are often recursive. It's possible, for example, to have `foldMap` for `T t` invoke `foldMap` for `T (Succ T)`. This means that each step in the recursion potentially risks building a whole `Foldable` dictionary just to pluck out its `foldMap`. Where that's an issue, it can be solved by either passing a function (instead of using a class constraint) or using a custom single-method class. Again: I don't know if we have an issue here; I just want to check.

treeowl:
I believe that the *unordered* folds and traversals actually have this problem, though I haven't pored over the Core to be sure. We can fix it easily using some extra classes like `FoldrU` and `FoldlU`.

treeowl:
Confirmed. When I compile  with `-ddump-simpl` and dig into the guts of `foldrU`, it eventually leads me to `foldr` for `BinomForest`. Cleaned up just a little bit, it looks like this:

```
Data.PQueue.Internals.$fFoldableBinomForest_$cfoldr
  :: forall (rk :: * -> *).
     Foldable rk =>
     forall a b. (a -> b -> b) -> b -> BinomForest rk a -> b
Data.PQueue.Internals.$fFoldableBinomForest_$cfoldr
  = \ (@ (rk_XpVJ :: * -> *))
      ($dFoldable1_XpVL :: Foldable rk_XpVJ)
      (@ a_apR9)
      (@ b_apRa) ->
      let {
        $dFoldable2_sqPe :: Foldable (Succ rk_XpVJ)
        $dFoldable2_sqPe
          = Data.PQueue.Internals.$fFoldableSucc
              @ rk_XpVJ $dFoldable1_XpVL } in
      let {
        lvl20_ssZO
          :: (a_apR9 -> b_apRa -> b_apRa)
             -> b_apRa -> BinomForest (Succ rk_XpVJ) a_apR9 -> b_apRa
        lvl20_ssZO
          = Data.PQueue.Internals.$fFoldableBinomForest_$cfoldr
[...]

PR #59, opened 2021-12-09
Streamline and expand folds

treeowl:
* As discussed in #55, we were recursively constructing `Foldable`
  dictionaries in unordered folds (for the basic queues, though not the
  `Prio` ones). Stop doing that.

* Add strict left unordered folds and monoidal unordered folds.

### DIFF
--- a/src/Data/PQueue/Internals.hs
+++ b/src/Data/PQueue/Internals.hs
@@ -1,4 +1,6 @@
-{-# LANGUAGE CPP, StandaloneDeriving #-}
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE StandaloneDeriving #-}
 
 module Data.PQueue.Internals (
   MinQueue (..),
@@ -34,8 +36,10 @@ module Data.PQueue.Internals (
   fromList,
   mapU,
   fromAscList,
+  foldMapU,
   foldrU,
   foldlU,
+  foldlU',
 --   traverseU,
   seqSpine,
   unions
@@ -47,6 +51,7 @@ import Data.Foldable (foldl')
 import Data.Semigroup (Semigroup(..), stimesMonoid)
 #endif
 
+import Data.PQueue.Internals.Foldable
 #ifdef __GLASGOW_HASKELL__
 import Data.Data
 import Text.Read (Lexeme(Ident), lexP, parens, prec,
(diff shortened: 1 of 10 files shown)


src/Data/PQueue/Internals/Foldable.hs
```diff
@@ -0,0 +1,22 @@
+-- | Writing 'Foldable' instances for nonuniform types in the natural manner
```
konsumlamm, line 1:
```suggestion
-- | Writing 'Foldable' instances for non-uniform types (e.g. 'BinomForest') in the natural manner
```
I assume by non-uniform you mean polymorphic recursive?

src/Data/PQueue/Prio/Internals.hs
```diff
@@ -650,6 +662,17 @@ foldrWithKeyF_ f fCh ts0 z0 = case ts0 of
     fCh' (Succ (BinomTree k a ts) tss) z =
       f k a (fCh ts (fCh tss z))
 
+-- | Unordered monoidal fold on a binomial forest.
+foldMapWithKeyF_ :: Monoid m => (k -> a -> m) -> (rk k a -> m) -> BinomForest rk k a -> m
```
konsumlamm, line 666: Why are you defining standalone functions here instead of also uisng a typeclass?

src/Data/PQueue/Internals.hs
```diff
@@ -648,12 +711,28 @@ mapU f (MinQueue n x ts) = MinQueue n (f x) (f <$> ts)
 -- | /O(n)/. Unordered right fold on a priority queue.
 foldrU :: (a -> b -> b) -> b -> MinQueue a -> b
 foldrU _ z Empty = z
-foldrU f z (MinQueue _ x ts) = x `f` foldr f z ts
+foldrU f z (MinQueue _ x ts) = x `f` foldr_ f z ts
 
--- | /O(n)/. Unordered left fold on a priority queue.
+-- | /O(n)/. Unordered left fold on a priority queue. This is rarely
+-- what you want; 'foldrU' and 'foldlU'' are more likely to perform
+-- well.
```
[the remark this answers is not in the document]
treeowl: True!

src/Data/PQueue/Internals/Foldable.hs
treeowl: I meant non-regular, sorry. Using non-regular (AKA nested) types tends to require polymorphic recursive functions. I'll fix the terminology.

src/Data/PQueue/Prio/Internals.hs
treeowl: Just going by the pattern the module was using before. I don't know which way (if either) performs better. We can evaluate that ... once we have benchmarks. Do you know of any benchmarks out there?

konsumlamm: I don't unfortunately, I've only done a few very basic benchmarks (`insert` and `deleteMin`) a while ago.

I think we should at least be consistent between `Data.PQueue` and `Data.PQueue.Prio`.

treeowl: Fine. I'll go with the class way for now and we can revisit later. I want to pack in a few more features (including deciding how to expose internals) and get a release out.

src/Data/PQueue/Internals.hs
```diff
@@ -603,6 +608,63 @@ instance Functor rk => Functor (BinomForest rk) where
   fmap f (Skip ts) = Skip (fmap f ts)
   fmap f (Cons t ts) = Cons (fmap f t) (fmap f ts)
 
+instance Foldr Zero where
+  foldr_ _ z ~Zero = z
```
konsumlamm, line 612: Btw, why are you using a lazy pattern match here?

src/Data/PQueue/Max.hs
```diff
@@ -273,10 +277,18 @@ mapU f (MaxQ q) = MaxQ (Min.mapU (\(Down a) -> Down (f a)) q)
 foldrU :: (a -> b -> b) -> b -> MaxQueue a -> b
 foldrU f z (MaxQ q) = Min.foldrU (flip (foldr f)) z q
 
+-- | /O(n)/. Unordered monoidal fold on a priority queue.
+foldMapU :: Monoid m => (a -> m) -> MaxQueue a -> m
+foldMapU f (MaxQ q) = Min.foldMapU (f . unDown) q
+
 -- | /O(n)/. Unordered left fold on a priority queue.
```
konsumlamm, line 284: You forgot the note about performance.

src/Data/PQueue/Internals.hs
```diff
@@ -645,12 +708,28 @@ mapU f (MinQueue n x ts) = MinQueue n (f x) (f <$> ts)
...
+-- what you want; 'foldrU' and 'foldlU'' are more likely to perform
+-- well.
 foldlU :: (b -> a -> b) -> b -> MinQueue a -> b
 foldlU _ z Empty = z
-foldlU f z (MinQueue _ x ts) = foldl f (z `f` x) ts
+foldlU f z (MinQueue _ x ts) = foldl_ f (z `f` x) ts
+
+-- | /O(n)/. Unordered strict left fold on a priority queue.
+--
+-- @since 1.4.2
```
konsumlamm, line 722: This seems to be the only file you added `@since` annotations. They should be on all new (exported) functions.

konsumlamm (commented):
I can only repeat what I said in https://github.com/lspitzner/pqueue/pull/46#pullrequestreview-825645049: please don't just merge non-trivial changes without review.

treeowl:
> I can only repeat what I said in [#46 (review)](https://github.com/lspitzner/pqueue/pull/46#pullrequestreview-825645049): please don't just merge non-trivial changes without review.

I thought I'd addressed your concerns. Let's try to keep this friendly.

src/Data/PQueue/Internals.hs
treeowl: Yes, you're right. I'll have to fix that up.

konsumlamm:
> I thought I'd addressed your concerns. Let's try to keep this friendly.

Sorry, I didn't mean to be unfriendly.

The person raising the concerns can usually judge best whether or not they've been addressed, so I think it's useful to await a re-review. Imo, a good rule for (non-controversial) PRs would be that it needs at least one explicit approval (from someone else than the author) to get merged.

src/Data/PQueue/Internals.hs
treeowl: There's no reason to actually match on the `Zero` constructor, since it carries no information. The pattern is only there to make it obvious what is there and why we don't care about it.

Issue #33, opened 2020-11-17
Offer untopped queues


treeowl:
`BinomHeap` can easily be wrapped up in a `newtype` to offer basic binomial queues that don't track size and don't store the minimal entry separately. For some applications, such as heapsort, this should be significantly more efficient. Let's do it.

treeowl:
I've done this now (for plain min-queues); there were no real difficulties. But as it is, there's a lot of code duplication. The fix, I think, is to define

```haskell
data MinQueue a = Empty | MinQueue {-# UNPACK #-} !Int !a !(BMinQueue a)
```

where `BMinQueue` is the "untopped" queue. For certain purposes, we'll need a conversion function from `BMinQueue` to `MinQueue` that extracts the minimal element and calculates the size. We could do those together, but I'm not sure it's worth the trouble. Even without that, I think a number of functions (partitions, say) will get faster, because it's more efficient to build a large untopped queue and then extract its minimum than to build a large topped queue. O(n) fewer comparisons, and probably also O(n) less allocation.

konsumlamm:
Hmm, I'm really not sure what to think of this one. On one hand, more efficient functions sound good, but otoh I'd really like to avoid the code duplication.

I don't think the average `pqueue` user would use such an untopped queue, nor do I think that heapsort is a common application (there are better sorting algorithms). So maybe this should be in an internal module or a separate package? However, that still doesn't avoid the code duplication...

konsumlamm:
Isn't a `BMinQueue` just a `BinomHeap` anyway? So we don't really need new types/modules if we just want more efficient operations.

konsumlamm:
Isn't a `BMinQueue` just a `BinomHeap` anyway? So we don't really need new types/modules if we just want more efficient operations.

treeowl:
The code duplication is gone. That was just an intermediate stage. I think the overall situation is much better now. Wouldn't this untopped version also be better for shortest path algorithms?

treeowl:
> Isn't a `BMinQueue` just a `BinomHeap` anyway? So we don't really need new types/modules if we just want more efficient operations.

Yes, a `BMinQueue` is a `BinomHeap`. It's an actual binomial queue, rather than one augmented with extra stuff. But giving it a newtype wrapper has at least two big benefits:

1. It can be treated as a proper abstract datatype.
2. We can give it typeclass instances.

[...]

Issue #21, opened 2020-06-06
Strict folds


konsumlamm:
There are currently no strict fold operations (except for the Foldable instances ofc) in this package. Would it be worthwhile to add strict variants of the different folds that are provided (like `foldrU`, `foldlU`)? Or is there any particular reason to not include them?

Regarding the implementation, they could just be implemented like the [default definitions](https://hackage.haskell.org/package/base-4.14.0.0/docs/src/Data.Foldable.html#Foldable) of `foldr'` and `foldl'`.

treeowl:
Yes. For min-queues, we want `foldlAsc'` and `foldrDesc'`. For max-queues, we want `foldlDesc'` and `foldrAsc'`. For all queues, we want `foldlU'` and possibly `foldrU'`. (Of course, we'll also want the `WithKey` versions.)

The current documentation for the unordered folds makes it pretty silly to have both left and right-handed versions. Obeying the documentation, we could define `foldlWithKeyU f = foldrWithKeyU (\k a b -> f b k a)`. We really should either:

1. Have `foldrU` and `foldlU'` only, or
2. Have all the unordered folds, but document a specific relationship between the orderings they use. For example, we might want to specify that `reverse . foldrU (:) [] = foldlU (flip (:)) []`. I think I prefer this option.

konsumlamm:
`foldlU'` was added in #59, so for unordered folds, this now has been addressed.

PR #64, opened 2021-12-10
Start working on bare queues

treeowl:
Addresses #33


treeowl:
@konsumlamm, this is still something of a work in progress, since I've only implemented plain (not `Prio`) min-queues. I'd like to get your thoughts on the general design before going any further.

treeowl:
@konsumlamm Could you take a look?

src/BinomialQueue/Max.hs
```diff
@@ -0,0 +1,292 @@
...
+-- License     :  BSD-style
+-- Maintainer  :  libraries@haskell.org
+-- Stability   :  experimental
+-- Portability :  portable
+--
+-- General purpose priority queue. Unlike the queues in "Data.PQueue.Max",
+-- these are /not/ augmented with a global root or their size, so 'getMax'
+-- and 'size' take logarithmic, rather than constant, time. When those
+-- operations are not (often) needed, these queues are generally faster than
+-- those in "Data.PQueue.Min".
```
konsumlamm, line 16:
```suggestion
-- those in "Data.PQueue.Max".
```

src/BinomialQueue/Max.hs
```diff
@@ -0,0 +1,292 @@
...
+
+-- | /O(log n)/. Returns the minimum element. Throws an error on an empty queue.
+findMax :: Ord a => MaxQueue a -> a
+findMax = fromMaybe (error "Error: findMax called on empty queue") . getMax
+
+-- | /O(1)/. The top (maximum) element of the queue, if there is one.
+getMax :: Ord a => MaxQueue a -> Maybe a
+getMax (MaxQueue q) = unDown <$> MinQ.getMin q
+
+-- | /O(log n)/. Deletes the minimum element. If the queue is empty, does nothing.
```
konsumlamm, line 120:
```suggestion
-- | /O(log n)/. Deletes the maximum element. If the queue is empty, does nothing.
```

src/BinomialQueue/Max.hs
```diff
@@ -0,0 +1,292 @@
...
+deleteMax :: Ord a => MaxQueue a -> MaxQueue a
+deleteMax = MaxQueue . MinQ.deleteMin . unMaxQueue
+
+-- | /O(log n)/. Extracts the minimum element. Throws an error on an empty queue.
```
konsumlamm, line 124:
```suggestion
-- | /O(log n)/. Extracts the maximum element. Throws an error on an empty queue.
```

src/BinomialQueue/Max.hs
```diff
@@ -0,0 +1,292 @@
...
+--
+-- If the order of the elements is irrelevant, consider using 'toListU'.
+toList :: Ord a => MaxQueue a -> [a]
+toList = fmap unDown . MinQ.toAscList . unMaxQueue
+
+toAscList :: Ord a => MaxQueue a -> [a]
+toAscList = fmap unDown . MinQ.toDescList . unMaxQueue
+
+toDescList :: Ord a => MaxQueue a -> [a]
+toDescList = fmap unDown . MinQ.toAscList . unMaxQueue
```
konsumlamm, line 211: These need doc comments. Same for a bunch of other functions below.

src/BinomialQueue/Min.hs
```diff
@@ -0,0 +1,222 @@
...
+  toAscList,
+  toDescList,
+  fromList,
+  fromAscList,
+  fromDescList,
+  -- * Unordered operations
+  mapU,
+  foldrU,
+  foldlU,
+  foldlU',
```
konsumlamm, line 80: In https://github.com/lspitzner/pqueue/issues/21#issuecomment-988157148 you argued for potentially only having `foldrU` and `foldlU'`, maybe we should do that for the new modules?

konsumlamm (commented):
Have you run the benchmarks to measure the performance improvements?

treeowl:
> Have you run the benchmarks to measure the performance improvements?

Yes. Most of them look good. One of them looks bad, and I haven't been able to figure out why. Regardless, I think the overall picture is good enough to make them worth exporting.

src/BinomialQueue/Min.hs
treeowl: Good point.

src/BinomialQueue/Max.hs
treeowl: Ugh, yes. Looks like I need lots of boring documentation work.

konsumlamm:
@treeowl any progress on this?

treeowl:
Paged out and busy. If you have time to finish it up, that would be awesome.

konsumlamm:
Upon further thought, I'm not so sure about the module names anymore. The fact that we're using binomial queues is an implementation detail that shouldn't matter to most users (and what if we decide to change the implementation to use skew-binomial heaps or trinomial heaps?), so I'm thinking maybe `Data.PQueue.Bare`/`PQueue.Bare` would be better after all?

> > Have you run the benchmarks to measure the performance improvements?
> 
> Yes. Most of them look good. One of them looks bad, and I haven't been able to figure out why. Regardless, I think the overall picture is good enough to make them worth exporting.

I ran the benchmarks myself and while I can't see any improvements for the `heapSort` benchmarks, most of the `kWay` benchmarks are about 5% faster (except one of them, as you mention, `1000 into 1000000 streams` is 9% slower). I'm not completely convinced that this is really worth it though.

I'm going to postpone this for now, to unblock further progress. I can rebase this PR later when needed.

treeowl:
If you want to unblock further progress elsewhere, why not just merge it and (temporarily, I hope) put in internal space?

konsumlamm:
> It's a more modular organization. You don't think that's worth it for itself?

Hmm, that's a good point actually. I'll merge this without exposing the new modules. https://github.com/lspitzner/pqueue/pull/64#discussion_r802079426 and https://github.com/lspitzner/pqueue/pull/64#discussion_r802087793 can be addressed later.

pqueue.cabal
```diff
@@ -35,9 +35,12 @@ library
     Data.PQueue.Prio.Max
     Data.PQueue.Min
     Data.PQueue.Max
+    BinomialQueue.Min
+    BinomialQueue.Max
   other-modules:
     Data.PQueue.Prio.Internals
     Data.PQueue.Internals
+    BinomialQueue.Internals
```
konsumlamm, line 43:
```suggestion
  other-modules:
    Data.PQueue.Prio.Internals
    Data.PQueue.Internals
    BinomialQueue.Internals
    BinomialQueue.Min
    BinomialQueue.Max
```

PR #65, opened 2021-12-16
Add k-way merge and heapsort benchmarks

treeowl:
Better to have some benchmarks than none.

### DIFF
--- a/benchmarks/BenchMinPQueue.hs
+++ b/benchmarks/BenchMinPQueue.hs
@@ -1,0 +1,36 @@
+import System.Random
+import Gauge
+
+import qualified KWay.PrioMergeAlg as KWay
+import qualified PHeapSort as HS
+
+kWay :: Int -> Int -> Benchmark
+kWay i n = bench
+  ("k-way merge looking " ++ show i ++ " deep into " ++ show n ++ " streams")
+  (whnf ((!! i) . KWay.merge . KWay.mkStreams n) $ mkStdGen 5466122035931067691)
+
+hSort :: Int -> Benchmark
+hSort n = bench
+  ("Heap sort with " ++ show n ++ " elements")
+  (nf (HS.heapSortRandoms n) $ mkStdGen (-7750349139967535027))
+
+main :: IO ()
+main = defaultMainWith defaultConfig{timeLimit = Just 15}
+  [ bgroup "heapSort"
+      [ hSort (10^3)
+      , hSort (10^4)
+      , hSort (10^5)
+      ]
+  , bgroup "kWay"
+      [ kWay (10^3) 1000000
+      , kWay (10^5) 1000
+      , kWay (10^5) 10000
+      , kWay (10^5) 100000
+      , kWay (10^6) 1000
+      , kWay (10^6) 10000
+      , kWay (10^6) 20000
+      , kWay (3*10^6) 1000
+      , kWay (2*10^6) 2000
+      , kWay (4*10^6) 100
+      ]
+  ]
--- a/benchmarks/BenchMinQueue.hs
+++ b/benchmarks/BenchMinQueue.hs
@@ -1,0 +1,36 @@
+import System.Random
+import Gauge
+
+import qualified KWay.MergeAlg as KWay
+import qualified HeapSort as HS
+
+kWay :: Int -> Int -> Benchmark
+kWay i n = bench
+  (show i ++ " into " ++ show n ++ " streams")
+  (whnf ((!! i) . KWay.merge . KWay.mkStreams n) $ mkStdGen 5466122035931067691)
+
+hSort :: Int -> Benchmark
+hSort n = bench
+  ("Heap sort with " ++ show n ++ " elements")
+  (nf (HS.heapSortRandoms n) $ mkStdGen (-7750349139967535027))
+
+main = defaultMainWith defaultConfig{timeLimit = Just 15}
+  [
+    bgroup "heapSort"
+      [ hSort (10^3)
+      , hSort (10^4)
+      , hSort (10^5)
+      ]
+  , bgroup "kWay"
+      [ kWay (10^3) 1000000
+      , kWay (10^5) 1000
+      , kWay (10^5) 10000
+      , kWay (10^5) 100000
+      , kWay (10^6) 1000
+      , kWay (10^6) 10000
+      , kWay (10^6) 20000
+      , kWay (3*10^6) 1000
+      , kWay (2*10^6) 2000
+      , kWay (4*10^6) 100
+      ]
+  ]
--- a/benchmarks/HeapSort.hs
+++ b/benchmarks/HeapSort.hs
@@ -1,0 +1,11 @@
+module HeapSort where
+
+import Data.PQueue.Min (MinQueue)
+import qualified Data.PQueue.Min as P
+import System.Random
+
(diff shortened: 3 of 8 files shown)


treeowl:
A couple results so far:

1. #26 improved all the results, some of them considerably.
2. If we arrange for specialization to the `Ord` instance, we can win maybe 20%.

treeowl:
@konsumlamm Do you have any objections to adding these few benchmarks, or can I just go ahead?

treeowl:
Wow, I just realized I did the k-way merge benchmark all wrong. I'll get that fixed!

benchmarks/KWay/MergeAlg.hs
```diff
@@ -0,0 +1,36 @@
+{-# language BangPatterns #-}
```
konsumlamm, line 1: I know that pragmas are case-insensitive, but I'd prefer a consistent approach (we're using `LANGUAGE` everywhere else).

benchmarks/BenchMinPQueue.hs
```diff
@@ -0,0 +1,37 @@
...
+kWay i n = bench
+  ("k-way merge looking " ++ show i ++ " deep into " ++ show n ++ " streams")
+  (whnf ((!! i) . KWay.merge . KWay.mkStreams n) $ mkStdGen 5466122035931067691)
+
+hSort :: Int -> Benchmark
+hSort n = bench
+  ("Heap sort with " ++ show n ++ " elements")
+  (nf (HS.heapSortRandoms n) $ mkStdGen (-7750349139967535027))
+
+main = defaultMain [
```
konsumlamm, line 17:
```suggestion
main :: IO ()
main = defaultMain [
```

konsumlamm (commented):
I have absolutely no objections to adding a few benchmarks, but some style nitpicks.

benchmarks/HeapSort.hs
```diff
@@ -0,0 +1,11 @@
+module HeapSort where
+import Data.PQueue.Min (MinQueue)
+import qualified Data.PQueue.Min as P
+import System.Random
+import Data.Int
+
+heapSortRandoms :: Int -> StdGen -> [Int32]
```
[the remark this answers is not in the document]
treeowl: It's an attempt to reduce the overhead of random number generation. I haven't checked to see that it works.

benchmarks/PHeapSort.hs
```diff
@@ -0,0 +1,11 @@
...
+import Data.PQueue.Prio.Min (MinPQueue)
+import qualified Data.PQueue.Prio.Min as P
+import System.Random
+import Data.Int
+
+heapSortRandoms :: Int -> StdGen -> [Int32]
+heapSortRandoms n gen = heapSort $ take n (randoms gen)
+
+heapSort :: Ord a => [a] -> [a]
+heapSort xs = [b | (b, ~()) <- P.toAscList . P.fromList . map (\a -> (a, ())) $ xs]
```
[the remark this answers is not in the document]
treeowl: I guess it shouldn't matter, because an outer `map` should fuse with `toAscList` (I *think* the latter has a rewrite rule) and avoid producing thunks to select the first components. But if that doesn't happen, things won't look great because `rnf` for lists doesn't fuse (maybe the assumption is that if you `rnf` something then you'll use it again, but that's not a great assumption).

benchmarks/KWay/MergeAlg.hs
treeowl: I'd prefer to make all the `language` pragmas lower case. I am absolutely sick of unnecessary shift key holding, and I don't have a capslock.

konsumlamm: Ok, fine.

benchmarks/BenchMinQueue.hs
```diff
@@ -0,0 +1,38 @@
...
+kWay i n = bench
+  (show i ++ " into " ++ show n ++ " streams")
+  (whnf ((!! i) . KWay.merge . KWay.mkStreams n) $ mkStdGen 5466122035931067691)
+
+hSort :: Int -> Benchmark
+hSort n = bench
+  ("Heap sort with " ++ show n ++ " elements")
+  (nf (HS.heapSortRandoms n) $ mkStdGen (-7750349139967535027))
+
+main = defaultMain [
```
konsumlamm, line 18:
```suggestion
main :: IO ()
main = defaultMain [
```

benchmarks/HeapSort.hs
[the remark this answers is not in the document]
treeowl: That turned out to be a bad idea. Fixed!

benchmarks/BenchMinQueue.hs
treeowl: Fixed.

treeowl:
OK, it should make a lot more sense now. Previously I was trying to merge *unsorted* streams, which was totally wrong and may well explain why the bare queues were performing worse than the augmented ones. Now the streams are sorted.

treeowl:
I'm struggling to get the time it takes to run the benchmarks down to something reasonable. I tried setting a time limit, but Gauge blew right past it by an order of magnitude. Clearly I'm doing something wrong. Is there some number of evaluations per sample I need to tweak somehow? I don't think we should delay merging to fix this, but we should get to it at some point.

konsumlamm (commented):
The time limit seems to work for me, any benchmark specifically that "blew right past it by an order of magnitude"?

We should consider using `tasty-bench` instead of `gauge`, it's a lot faster (and much lighter on dependencies), it wouldn't even require a time limit, although sometimes `kWay (10^3) 1000000` seems to get stuck, maybe we should just remove that bench. Also, `tasty-bench` is much more actively maintained.

benchmarks/KWay/RandomIncreasing.hs
```diff
@@ -0,0 +1,24 @@
...
+data Stream = Stream !Word64 {-# UNPACK #-} !StdGen
+
+viewStream :: Stream -> (Word64, Stream)
+viewStream (Stream w gen) = (w, case uniform gen of (k, gen') -> Stream (w + fromIntegral (k :: Word16)) gen')
+
+mkStream :: StdGen -> (Stream, StdGen)
+mkStream gen
+  | (gen1, gen2) <- split gen
+  , (w, gen1') <- uniform gen1
+  = (Stream w gen1', gen2)
```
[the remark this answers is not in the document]
treeowl: Yes, of course you're right.

treeowl:
> The time limit seems to work for me, any benchmark specifically that "blew right past it by an order of magnitude"?
> 
> We should consider using `tasty-bench` instead of `gauge`, it's a lot faster (and much lighter on dependencies), it wouldn't even require a time limit, although sometimes `kWay (10^3) 1000000` seems to get stuck, maybe we should just remove that bench. Also, `tasty-bench` is much more actively maintained.

I ran into trouble when I added heapsort benchmarks for longer lists. 10^6 elements was unbearably slow. I don't think I ever got 10^7 to finish benchmarking. I don't really care what benchmarking framework we use, as long as it's reliable. Should we merge this, and then let you replace it with tasty? I don't know how to set that up.

konsumlamm (approved):
> I don't really care what benchmarking framework we use, as long as it's reliable. Should we merge this, and then let you replace it with tasty? I don't know how to set that up.

Ok (it's really straightforward).

PR #68, opened 2021-12-17
Remove redundant `Control.Applicative.Identity`

konsumlamm:
Use `Data.Functor.Identity` from `base` instead.


treeowl:
Oh good. I didn't know when that reached `base`.

treeowl (approved):

konsumlamm:
> Oh good. I didn't know when that reached `base`.

It was added in 4.8, which is exactly our lower bound for `base`.

PR #69, opened 2021-12-17
Small updates

konsumlamm:


treeowl:
Could you look at that redundant import in the test code while you're at it?

pqueue.cabal
```diff
@@ -69,14 +69,9 @@ test-suite test
...
   , pqueue
   }
   ghc-options: {
     -Wall
   }
-  if impl(ghc >= 8.0) {
-    ghc-options: {
-      -fno-warn-unused-imports
-    }
-  }
```
treeowl, line 82: Yay!

treeowl (approved):

konsumlamm:
> Could you look at that redundant import in the test code while you're at it?

Thanks for reminding me! The solution was to update the lower bound for `QuickCheck` to 2.10 (`applyFun2` was added there, so previous versions wouldn't work anyway).

PR #71, opened 2021-12-18
Use `tasty-bench` instead of `gauge`

konsumlamm:
Reasons for using `tasty-bench`:
* actively maintained
* works on GHC 9.2
* [much more lightweight than `gauge`](https://github.com/Bodigrim/tasty-bench#how-lightweight-is-it) (less dependencies, less code, faster to compile)
* tests run a lot faster
* builtin support for [baseline comparisons](https://github.com/Bodigrim/tasty-bench#comparison-against-baseline)

Also see the [description](https://github.com/Bodigrim/tasty-bench#readme).

### DIFF
--- a/.github/workflows/haskell.yaml
+++ b/.github/workflows/haskell.yaml
@@ -19,3 +19,5 @@ jobs:
         run: cabal build
       - name: Test
         run: cabal test --test-show-details=direct
+      - name: Build benchmarks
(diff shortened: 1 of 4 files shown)


treeowl (approved):
LGTM. I'm going to check my `bare` branch against the (hopefully correct, now) benchmarks and assuming I like what I see I'd like to get that merged pretty quickly.

PR #79, opened 2022-04-17
Expose `Data.PQueue.Prio.Max.foldlU'`

konsumlamm:
This was added in #59, but only to `Data.PQueue.Prio.Max.Internals`, without being exposed by `Data.PQueue.Prio.Max`. @treeowl I assume this was an oversight?


konsumlamm:
@treeowl can you confirm that this was indeed an oversight?

treeowl:
Probably? I don't quite remember.

Issue #78, opened 2022-04-17
Two equal priority queues with keys can compare as unequal


konsumlamm:
Two equal queues may compare as unequal, since duplicate keys are allowed and the `Eq` implementation takes only care of comparing the keys in order.

**Example:**
```hs
ghci> import qualified Data.PQueue.Min as PMin
ghci> PMin.fromAscList [(1, 'a'), (1, 'b')] == PMin.fromAscList [(1, 'b'), (1, 'a')]
False
```

This should be documented, as I don't think this can be fixed without making the implementation significantly slower.

treeowl:
Hrmmmm..... Which operations don't respect `==`?

konsumlamm:
> Hrmmmm..... Which operations don't respect `==`?

I'm not sure what you mean, the issue is that entries with equal keys may be in any order.

treeowl:
The usual expectation is that if

```haskell
f :: A -> B
x, y :: A
x == y = True
```

then

```haskell
f x == f y = True
```

We say that `f` *respects* `==`.

Typically, "safe" API functions should respect `==`. If there's a relatively small and well defined set of operations that don't, that can be acceptable for the sake of practicality (see `HashMap`, for example, where two maps may compare equal even though converting them both to lists may produce differently ordered lists and they may be traversed in different orders). But if it turns out that the most essential priority queue operations don't respect `==`, then that `==` will break everyone's intuition. In that case, it should preferably be changed, or if there's no way to do anything sensible then the `Eq` instance should likely be removed.

We should start by looking at how insertion and view interact with `==`, since those are the most important operations.

treeowl:
Similarly for binary operations. Given

```haskell
a == a' = True
b == b' = True
```

users will want to be able to rely on

```haskell
merge a b == merge a' b' = True
```

My bet is that the only ways to make sensible and abstract `Eq` instances for the `Prio` queues are to consider the results of `minView` (hence `toList`) and/or `merge` non-deterministic. Considering only `minView` non-deterministic is easy:

```haskell
a == b = sort (toList a) == sort (toList b)
```

I believe we currently consider only `merge` non-deterministic in our `Eq` instance, which leads to very difficult/brittle code. So I think we should to the simple thing instead, if we want an `Eq` instance.

konsumlamm:
[...]

PR #81, opened 2022-04-23
Update tests

### LINKED ISSUE #49: Test all queue types and functions
We mostly just test min-queues. We should be testing every function and method with a nontrivial implementation for every queue type!

konsumlamm:
Resolves #49.

The test suite now tests all of the four priority queue types. The tests are pretty repetitive, but I sadly don't see a good way to avoid that. The key-value priority queue tests use `()` as the value type, to avoid the problems of #78.

### DIFF
--- a/pqueue.cabal
+++ b/pqueue.cabal
@@ -52,19 +52,16 @@ library
   other-extensions:
       BangPatterns
     , CPP
-  ghc-options: {
+  ghc-options:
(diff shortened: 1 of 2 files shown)


konsumlamm:
@treeowl could you take a look at this?

konsumlamm:
@treeowl do you want to take a look or can I just merge this?

PR #90, opened 2023-03-16
bump base upper bound to <4.19

mitchellwrosen:


mitchellwrosen:
CI seems busted. 😢

treeowl:
We'll have to see if it recovers or if we need to regenerate with `haskell-ci` to catch up with whatever.

mitchellwrosen:
Sounds good. I went ahead and added another commit here just to see what this looks like with a `haskell-ci github pqueue.cabal`-generated CI config. We can remove that commit before merging if we'd like.

treeowl:
We already have a Haskell CI workflow. Let's see what CI does this time, but if the new one works, you should revert that commit and use `haskell-ci regenerate` instead to replace the old one instead of adding another.

mitchellwrosen:
Sure; I'll just revert the commits then. I saw the existing CI file, but it looked hand-written, and I thought `haskell-ci` would be an improvement.

Feel free to add more commits to this PR, or not - I'm stepping away from it for now.

treeowl:
Sorry if I got it wrong. I was just trying to advise. It looks like you did a good thing....

treeowl:
It looks like I did get it wrong.  If that was from haskell-ci, it was way too old. Ugh. Sorry.

mitchellwrosen:
I think this PR is good to merge; here's one green run from history: https://github.com/lspitzner/pqueue/actions/runs/4433666369

To fix CI I simply removed `7.10.3` from `Tested-With`, as it seems ghcup has trouble installing that one. But that can be another PR.

treeowl:
Okay, thanks a lot. I'll try to remember to do the CI thing and make a release in the morning.

mitchellwrosen:
No problem, thank you!

treeowl:
It looks like whatever problem you encountered with 7.10.3 was temporary (maybe a server issue or something). I didn't have any trouble with it in #91, so I guess we'll keep supporting 7.10.3 a bit longer.

konsumlamm:
> It looks like whatever problem you encountered with 7.10.3 was temporary (maybe a server issue or something). I didn't have any trouble with it in #91, so I guess we'll keep supporting 7.10.3 a bit longer.

I think the issue is that installing 7.10.3 via GHCup no longer works (by default), which is what the old CI used (via `haskell/actions/setup`). The new CI uses `hvr-ppa` instead.

ocharles:
Could we get this released to Hackage, or at least a Hackage revision?

treeowl:
Yeah, let's just make a release and deal with the remaining equality question later.

ocharles:
The what? I don't see any mention of equality here?

konsumlamm:
The equality question is https://github.com/lspitzner/pqueue/issues/78. See also https://github.com/lspitzner/pqueue/issues/121.

PR #92, opened 2023-03-17
Pattern synonyms and Data instances

### LINKED ISSUE #50: Data instances are unsafe/wrong

treeowl:
* Add some pattern synonyms.
* Improve the `Data` instances to
  - Make the `MinPQueue` version incremental, like the `MinQueue` one.
  - Make all the instances respect the queue invariants.
* Use the pattern synonyms to explain the `Data` instances.

Fixes #50.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
     ([#92](http://github.com/lspitzner/pqueue/pull/92))
 
   * Make the `Data` instances respect the queue invariants. Make the
-    `Constructor`s match the pattern synonyms. Make the `Data` instance for
+    `Constr`s match the pattern synonyms. Make the `Data` instance for
     `MinPQueue` work "incrementally", like the one for `MinQueue`.
     ([#92](http://github.com/lspitzner/pqueue/pull/92))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Revision history for pqueue
 
+## 1.5.0
+
+  * Add pattern synonyms to work with `MinQueue` and `MinPQueue`.
+    ([#92](http://github.com/lspitzner/pqueue/pull/92))
+
+  * Make the `Data` instances respect the queue invariants. Make the
+    `Constructor`s match the pattern synonyms. Make the `Data` instance for
+    `MinPQueue` work "incrementally", like the one for `MinQueue`.
+    ([#92](http://github.com/lspitzner/pqueue/pull/92))
+
 ## 1.4.3.0 -- 2022-10-30
 
   * Add instances for [indexed-traversable](https://hackage.haskell.org/package/indexed-traversable).
--- a/src/Data/PQueue/Internals.hs
+++ b/src/Data/PQueue/Internals.hs
@@ -1,8 +1,15 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE BangPatterns #-}
+#ifdef __GLASGOW_HASKELL__
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE ViewPatterns #-}
+#endif
 
 module Data.PQueue.Internals (
   MinQueue (..),
+#ifdef __GLASGOW_HASKELL__
+  pattern (:<),
+#endif
   BinomHeap,
   BinomForest(..),
   BinomTree(..),
@@ -81,6 +88,24 @@ fromBare xs = case BQ.minView xs of
   Nothing -> Empty
 
 #ifdef __GLASGOW_HASKELL__
+infixr 5 :<
+
+-- | A bidirectional pattern synonym for working with the minimum view of a
+-- 'MinPQueue'. Using @:<@ to construct a queue performs an insertion.
+--
+-- @since 1.5.0
+#  if __GLASGOW_HASKELL__ >= 800
+pattern (:<) :: Ord a => a -> MinQueue a -> MinQueue a
+#  else
+pattern (:<) :: () => Ord a => a -> MinQueue a -> MinQueue a
+#  endif
+pattern a :< q <- (minView -> Just (a, q))
+  where
+    a :< q = insert a q
+
+-- | Treats the priority queue as an empty queue or a minimal element and a
+-- priority queue. The constructors, conceptually, are 'Empty' and '(:<)'. All
+-- constructed queues maintain the queue invariants.
 instance (Ord a, Data a) => Data (MinQueue a) where
   gfoldl f z q = case minView q of
     Nothing      -> z Empty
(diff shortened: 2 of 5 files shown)


treeowl:
@konsumlamm I suspect we'll want similar pattern synonyms for max-queues, but do we want

```haskell
(:>) :: Ord a => MaxQueue a -> a -> MaxQueue a
```

or

```haskell
(<:) :: Ord a => a -> MaxQueue a -> MaxQueue a
```

I think the first option is clearer, since I think of the maximum as coming *after* the rest. But ergonomics will then be quite different between min-queues and max-queues, and I'm not sure how good that is. Hmmmmm....

Semi-separately: now that the `Ord` instance for `Data.Ord.Down` is decent, we should switch to it. We could then export `MaxQueue` and `MaxPQueue` concretely (with their constructors). Should they then use *derived* `Data` instances? I think that probably makes the most sense, but I'm not sure.

treeowl:
@konsumlamm In case you didn't notice, I was hoping for a review of this.

konsumlamm:
> @konsumlamm In case you didn't notice, I was hoping for a review of this.

I did notice, don't worry. It just takes some time to get into the issue again and I still have other things to do.

src/Data/PQueue/Internals.hs
```diff
@@ -81,15 +81,20 @@ fromBare xs = case BQ.minView xs of
...
 instance (Ord a, Data a) => Data (MinQueue a) where
   gfoldl f z q = case minView q of
     Nothing      -> z Empty
     Just (x, q') -> z insert `f` x `f` q'
 
   gunfold k z c = case constrIndex c of
     1 -> z Empty
-    2 -> k (k (z insertMinQ))
-    _ -> error "gunfold"
+    2 -> k (k (z insert))
```
konsumlamm, line 96: Is the double `k` correct? It looks a bit suspicious to me.

konsumlamm (commented):
Again, I don't know much about `Data`, so I'm just looking if things feel right.

konsumlamm:
I'm reluctant about the pattern synonyms, as they're not O(1) (while standard patterns are) and people might expect them to be, unknowingly introducing a slowdown.

Is there a problem with keeping the `Constr`s as is?

konsumlamm:
Btw, if you have time, could you take a look if #55 can be closed?

src/Data/PQueue/Prio/Internals.hs
```diff
@@ -80,22 +80,37 @@ build f = f (:) []
...
+
+  dataCast1 x = gcast1 x
+
+  toConstr q
+    | null q = emptyConstr
+    | otherwise = consConstr
+
   dataTypeOf _ = queueDataType
-  dataCast1 f  = gcast1 f
-  dataCast2 f  = gcast2 f
```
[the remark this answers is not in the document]
treeowl: Definitely a mistake. Thanks for catching it.

treeowl:
@konsumlamm I think I've addressed all the issues. I would like to squash the commits rather than merging them.

treeowl:
> Is there a problem with keeping the `Constr`s as is?

I really don't like the inconsistency between the way the plain versions and `Prio` versions conceptualize their types. I also think it's much easier to "explain" the `Data` instances by reference to bidirectional patterns than by reference to plain functions (one of which doesn't even bother to exist). Since the `Data` instance for `MinQueue` was fairly broken from the jump, I'd be pretty surprised if anyone actually used them in any way that the proposed changes would break.

konsumlamm:
Hmm, ok.

As for your question about `:>` vs `<:`, I don't really have an opinion on that. I'm fine with using `:>`, as that's similar to what `Data.Sequence` does.

Do you also want to improve the `Data` instances for max queues in this PR or leave that to a future PR?

treeowl:
I don't want to make any decisions whatsoever about max queues in this PR. I'll open a new one for pattern synonyms for them.

konsumlamm:
> I would like to squash the commits rather than merging them intact.

That's what I always do anyway.

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Revision history for pqueue
 
+## 1.5.0
+
+  * Add pattern synonyms to work with `MinQueue` and `MinPQueue`.
+    ([#92](http://github.com/lspitzner/pqueue/pull/92))
+
+  * Make the `Data` instances respect the queue invariants. Make the
+    `Constructor`s match the pattern synonyms. Make the `Data` instance for
```
konsumlamm, line 9:
```suggestion
    `Constr`s match the pattern synonyms. Make the `Data` instance for
```

treeowl:
Are you ready to approve?

konsumlamm (approved):

Issue #56, opened 2021-12-08
Switch to Data.Ord.Down


treeowl:
We don't really want our own version. We should make the switch once https://github.com/haskell/core-libraries-committee/issues/23 is accepted and the change is made.

konsumlamm:
If we're waiting for haskell/core-libraries-committee#23, the question arises if we only want to use `Data.Ord.Down` for newer versions (for which these optimizations are inplemented). My preference is to use it anyway, since it's a lot cleaner imo (and people who really need the performance can use a newer GHC).

treeowl:
It's a separate module, so no big deal to keep it around for a while and only compile it when necessary. We can toss it a couple years later. Of course, that's all assuming my proposal is accepted....

PR #93, opened 2023-03-17
Use Data.Ord.Down

treeowl:
We used to use our own `Down` newtype because the one in `Data.Ord` had a somewhat lousy `Ord` instance. That has since been corrected, so we can just do what everyone else does.


treeowl:
Hmmm.... That's a lot of mess still. We presumably don't want to expose orphan instances from `base-orphans`. For some things (like `Foldable`) we could do the gunk by hand. For others, like `Data`, I imagine the best we can do is some `FlexibleContexts` junk with constraints like `Data (Down a)`, letting downstream pull in `base-orphans`. Not the prettiest situation at all. Perhaps the whole `Down` swap should go on the back burner for a few more years?

konsumlamm:
> Perhaps the whole `Down` swap should go on the back burner for a few more years?

Yeah, I think that's the best option for now. Having constraints on `Down` leaks internal details (the users shouldn't have to care about wether we use `Down` or something else) and implementing the instances by hand defeats the whole point of this change (namely to reduce code size). Using `base-orphans` would be an option, but that also pulls in a whole bunch of other instances (that might even conflict with some local orphan instances), so not a great option.

Let's also link #56, as that's the issue for this.

One thing I did notice though, we can probably replace a bunch of the `fmap Down` and `fmap unDown` by `coerce`, since `Down` is a newtype (`Down` and `unDown` too, but that shouldn't make a difference). We should also add benchmarks for max queues, to examine their overhead over min queues.

PR #95, opened 2023-03-26
Deprecate seqSpine

treeowl:
The `seqSpine` function was useful back in the day when all sorts of operations were written in a way that allowed thunks to accumulate along the spine. Now, we carefully force as much of the spine as the amortized bounds allow, which eliminates this problem in general. The only operations that still accumulate thunks along the spine are monotonic maps, and forcing the spines of those just leads to the creation of more thunks at the leaves—it's not really useful.

### DIFF
--- a/src/Data/PQueue/Max.hs
+++ b/src/Data/PQueue/Max.hs
@@ -371,8 +371,9 @@ keysQueue (Prio.MaxPQ q) = MaxQ (Min.keysQueue q)
 --
 -- Note: The spine of a 'MaxQueue' is stored somewhat lazily. Most operations
 -- take great care to prevent chains of thunks from accumulating along the
--- spine to the detriment of performance. However, 'mapU' can leave expensive
--- thunks in the structure and repeated applications of that function can
--- create thunk chains.
+-- spine to the detriment of performance. The only one that doesn't,
+-- 'mapU', does not benefit from forcing the spine anyway. Forcing the spine
+-- of the result just forces the creation of thunks for all the elements.
+{-# DEPRECATED seqSpine "This function is no longer useful." #-}
 seqSpine :: MaxQueue a -> b -> b
 seqSpine (MaxQ q) = Min.seqSpine q
--- a/src/Data/PQueue/Prio/Max/Internals.hs
+++ b/src/Data/PQueue/Prio/Max/Internals.hs
@@ -577,10 +577,12 @@ toListU (MaxPQ q) = fmap (first' unDown) (Q.toListU q)
 
 -- | \(O(\log n)\). @seqSpine q r@ forces the spine of @q@ and returns @r@.
 --
--- Note: The spine of a 'MaxPQueue' is stored somewhat lazily. Most operations
+-- Note: The spine of a 'MinPQueue' is stored somewhat lazily. Most operations
 -- take great care to prevent chains of thunks from accumulating along the
--- spine to the detriment of performance. However, 'mapKeysMonotonic' can leave
--- expensive thunks in the structure and repeated applications of that function
--- can create thunk chains.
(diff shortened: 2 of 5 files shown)


src/Data/PQueue/Max.hs
```diff
@@ -371,8 +371,8 @@ keysQueue (Prio.MaxPQ q) = MaxQ (Min.keysQueue q)
 --
 -- Note: The spine of a 'MaxQueue' is stored somewhat lazily. Most operations
 -- take great care to prevent chains of thunks from accumulating along the
--- spine to the detriment of performance. However, 'mapU' can leave expensive
--- thunks in the structure and repeated applications of that function can
--- create thunk chains.
+-- spine to the detriment of performance. The only one that doesn't,
+-- 'mapKeysMonotonic', does not benefit from forcing the spine anyway.
```
konsumlamm, line 375:
```suggestion
-- 'mapU', does not benefit from forcing the spine anyway.
```
I think you confused `Data.PQueue.Max` and `Data.PQueue.Prio.Max` here.

src/Data/PQueue/Prio/Max/Internals.hs
```diff
@@ -579,8 +579,8 @@ toListU (MaxPQ q) = fmap (first' unDown) (Q.toListU q)
 --
 -- Note: The spine of a 'MaxPQueue' is stored somewhat lazily. Most operations
 -- take great care to prevent chains of thunks from accumulating along the
--- spine to the detriment of performance. However, 'mapKeysMonotonic' can leave
--- expensive thunks in the structure and repeated applications of that function
--- can create thunk chains.
+-- spine to the detriment of performance. The only one that doesn't, 'mapU',
+-- does not benefit from forcing the spine anyway.
```
konsumlamm, line 583:
```suggestion
-- spine to the detriment of performance. The only one that doesn't,
-- 'mapKeysMonotonic', does not benefit from forcing the spine anyway.
```
I think you confused `Data.PQueue.Max` and `Data.PQueue.Prio.Max` here.

konsumlamm:
Can you elaborate why `mapU`/`mapKeysMonotonic` don't benefit from `seqSpine`? Is it because all the work is accumulated in the elements?

treeowl:
> Can you elaborate why `mapU`/`mapKeysMonotonic` don't benefit from `seqSpine`? Is it because all the work is accumulated in the elements?

Exactly. I've updated the documentation to explain that, and I've also added references to more functions with similar issues in he `Prio` case.

treeowl:
Mapping over a queue gives you (a thunk for) the first element, and a thunk for the rest of the queue. As you push your way down the spine, successive vertebrae each produce larger and larger numbers of thunks.

src/Data/PQueue/Prio/Max/Internals.hs
```diff
@@ -577,10 +577,12 @@ toListU (MaxPQ q) = fmap (first' unDown) (Q.toListU q)
 
 -- | \(O(\log n)\). @seqSpine q r@ forces the spine of @q@ and returns @r@.
 --
--- Note: The spine of a 'MaxPQueue' is stored somewhat lazily. Most operations
+-- Note: The spine of a 'MinPQueue' is stored somewhat lazily. Most operations
```
konsumlamm, line 580: This was correct, wasn't it?

treeowl: Ugh. Whoops.

treeowl:
Hahaha. This entire PR is based on me getting confused. Whooops. There really was a (minor) reason for these functions to persist, and it actually was *only* the functions that were in the documentation. I should explain more about *why* that is somewhere...

Issue #35, opened 2020-11-18
Get rid of insertBehind


treeowl:
It's extremely slow, breaks on `union`, and relies in a brittle fashion on the exact implementation of `minView`. Let's be rid of it! The operation makes much more sense in priority queues based on finger trees, where the elements are stored in insertion order and the structure is annotated with subtree minima. (Specifically, a 2-3 finger tree whose digits can be 0 or 1 should do the trick).

treeowl:
Serious question: does this function even work at the moment? I have no idea if the new deletion implementation satisfies its brittle needs. If not, we may need to make a make a major release next just so we can kill this thing. It should never have gone in to begin with.

treeowl:
Maybe I've been unfair. I thought this broke on union, but maybe it doesn't. And I haven't been able to break it (yet) with the new deletion function. So... maybe it's okay? It's really slow though!

treeowl:
I just don't like how brittle it is. Every single operation in the module has to be written to preserve the order of `insertBehind` even though the module header explicitly documents unstable ordering. If it were *fast*, that might be worth doing (for stable sorts), but it's not, so I'd much rather not have that landmine.

PR #96, opened 2023-03-27
Deprecate insertBehind

treeowl:
`insertBehind` does not interact in any particularly sensible way with merges. I don't have a lot of faith that it works in other cases at all—it's far too fragile. It's also too slow to be really useful.

First stage in #35

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## 1.5.0
 
+  * Deprecate `insertBehind`. This function does not play nicely with merges,
+    we lack tests to verify it works properly without merges, it imposes a
+    substantial maintenance burden on the rest of the package, and it is quite
+    slow. ([#35](https://github.com/lspitzner/pqueue/issues/35))
+
   * Add pattern synonyms to work with `MinQueue` and `MinPQueue`.
     ([#92](http://github.com/lspitzner/pqueue/pull/92))
 
--- a/src/Data/PQueue/Prio/Internals.hs
+++ b/src/Data/PQueue/Prio/Internals.hs
@@ -272,6 +272,7 @@ insert k a (MinPQ n k' a' ts)
 -- Insert an element with the specified key into the priority queue,
 -- putting it behind elements whose key compares equal to the
 -- inserted one.
+{-# DEPRECATED insertBehind "This function is not reliable." #-}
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## 1.5.0
 
+  * Deprecate `insertBehind`. This function does not play nicely with merges, I
+    don't have a lot of faith that it works properly without merges, and it is
+    quite slow. ([#35](https://github.com/lspitzner/pqueue/issues/35))
```
konsumlamm, line 7:
```suggestion
  * Deprecate `insertBehind`. This function does not play nicely with merges, I
    don't have a lot of faith that it works properly without merges, and it is
    quite slow. ([#35](https://github.com/lspitzner/pqueue/issues/35))
```
IMO changelogs should be a bit more formal (i.e. no "I don't have a lot of faith").

konsumlamm (approved):
Seems fine to me. It's weird to have a function that makes guarantees about the order of elements with equal keys, when all the other functions don't. I guess in most cases `insert` would suffice. At least no package on Hackage seems to use it (https://hackage-search.serokell.io/?q=insertBehind). I'm reluctant to entirely remove it any time soon though.

CHANGELOG.md
treeowl: I can be more formal. Suggestions?

treeowl:
> I'm reluctant to entirely remove it any time soon though.

We've never had any tests to demonstrate that it works. Those tests will not be easy, since other operations in the module could potentially mess up the relative ordering. I don't want to support this function, and I want it gone ASAP. It belongs in a different data structure altogether—one with two "fingers" rather than one.

treeowl:
To clarify what I said in my last comment, that one function puts a major maintenance burden on other functions (especially `insert` and `minViewWithKey`) without contributing much on its own. It also has no specified (or obvious) merge semantics, so it can't be used usefully with merges (which are the raison d'être of binomial queues). It's very expensive dead weight, and we don't even know if we've met the heavy maintenance burden.

konsumlamm:
I think we don't have to consider deprecated functions too much when changing other functions. Just provide it as "it's there, maybe it works". I'm not against removing it at all, just not that soon.

PR #99, opened 2023-03-29
Remove `foldWhileFB`

### LINKED ISSUE #98: Remove foldWhileFB in favor of Data.List.takeWhile?
We currently implement `takeWhile` in terms of a funny `foldWhileFB` function, which Louis Wasserman commented was a better producer than `Data.List.takeWhile`. That was in 2015. The code for implementing list fusion of `Data.List.takeWhile` was written in 2014 (and therefore may not have been released when Louis wrote that comment). Additionally, Takano Akio add an inlining pragma to make that work better in 2017. As a result of all these changes, I suspect we can probably just use `Data.List.takeWhile`, which would be a rather nicer thing to do.

konsumlamm:
Closes #98.

`takeWhile` has a rewrite rule that does the same (https://hackage.haskell.org/package/base-4.18.0.0/docs/src/GHC.List.html#takeWhile) since base-4.9. I also did a small micro-benchmark that seems to confirm that there is no performance regression.

### DIFF
--- a/src/BinomialQueue/Max.hs
+++ b/src/BinomialQueue/Max.hs
@@ -136,7 +136,7 @@ maxView (MaxQueue q) = case MinQ.minView q of
 (!!) :: Ord a => MaxQueue a -> Int -> a
 q !! n  | n >= size q
     = error "BinomialQueue.Max.!!: index too large"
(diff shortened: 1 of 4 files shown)


treeowl (approved):
Yasssss.

PR #102, opened 2023-03-30
Strict maps

### LINKED ISSUE #60: Add strict mapping functions
Since the queues are mostly strict, mapping in general is pretty slow, but lazy maps are egregiously so: they fill the structure with one thunk per element. Let's add strict versions that at least *accomplish* something for the work they do.

treeowl:
Closes #60

### DIFF
--- a/src/Data/PQueue/Prio/Internals.hs
+++ b/src/Data/PQueue/Prio/Internals.hs
@@ -24,6 +24,7 @@ module Data.PQueue.Prio.Internals (
   updateMinWithKeyA',
   minViewWithKey,
   mapWithKey,
+  mapWithKey',
   mapKeysMonotonic,
   mapMaybeWithKey,
   mapEitherWithKey,
@@ -43,8 +44,10 @@ module Data.PQueue.Prio.Internals (
   foldlWithKeyU,
   foldlWithKeyU',
   traverseWithKey,
+  traverseWithKey',
   mapMWithKey,
   traverseWithKeyU,
+  traverseWithKeyU',
   seqSpine,
   mapForest,
   unions
@@ -341,6 +344,12 @@ minViewWithKey (MinPQ n k a ts) = Just ((k, a), extractHeap n ts)
 -- | \(O(n)\). Map a function over all values in the queue.
 mapWithKey :: (k -> a -> b) -> MinPQueue k a -> MinPQueue k b
 mapWithKey f = runIdentity . traverseWithKeyU (Identity .: f)
+
+-- | \(O(n)\). Map a function over all values in the queue, forcing the results.
+--
+-- @since 1.5.0
+mapWithKey' :: (k -> a -> b) -> MinPQueue k a -> MinPQueue k b
+mapWithKey' f = runIdentity . traverseWithKeyU' (Identity .: f)
 
 -- | \(O(n)\). @'mapKeysMonotonic' f q == 'mapKeys' f q@, but only works when @f@ is strictly
 -- monotonic. /The precondition is not checked./ This function has better performance than
@@ -688,6 +697,15 @@ traverseWithKey f q = case minViewWithKey q of
   Nothing      -> pure empty
   Just ((k, a), q')  -> liftA2 (insertMin k) (f k a) (traverseWithKey f q')
 
+-- | A version of 'traverseWithKey' that forces all the results before
+-- installing them in a queue.
+--
+-- @since 1.5.0
+traverseWithKey' :: (Ord k, Applicative f) => (k -> a -> f b) -> MinPQueue k a -> f (MinPQueue k b)
+traverseWithKey' f q = case minViewWithKey q of
+  Nothing      -> pure empty
+  Just ((k, a), q')  -> liftA2 (insertMin k) (f k a) (traverseWithKey' f q')
+
 -- | A strictly accumulating version of 'traverseWithKey'. This works well in
 -- 'IO' and strict @State@, and is likely what you want for other "strict" monads,
 -- where @⊥ >>= pure () = ⊥@.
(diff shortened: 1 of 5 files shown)


src/Data/PQueue/Prio/Internals.hs
```diff
@@ -688,6 +697,15 @@ traverseWithKey f q = case minViewWithKey q of
...
 
+-- | A version of 'traverseWithKey' that forces all the results before
+-- installing them in a queue.
+--
+-- @since 1.5.0
+traverseWithKey' :: (Ord k, Applicative f) => (k -> a -> f b) -> MinPQueue k a -> f (MinPQueue k b)
+traverseWithKey' f q = case minViewWithKey q of
+  Nothing      -> pure empty
+  Just ((k, a), q')  -> liftA2 (insertMin k) (f k a) (traverseWithKey' f q')
+
```
konsumlamm, line 708: This doesn't look like it forces anything, it's the same as `traverseWithKey`.

src/Data/PQueue/Prio/Internals.hs
```diff
@@ -709,6 +727,16 @@ traverseWithKeyU :: Applicative f => (k -> a -> f b) -> MinPQueue k a -> f (MinP
...
 
+-- | \(O(n)\). An unordered traversal over a priority queue, in no particular order.
+-- While there is no guarantee in which order the elements are traversed, the resulting
+-- priority queue will be perfectly valid. The results are forced before they are installed
+-- in the queue.
+--
+-- @since 1.5.0
+traverseWithKeyU' :: Applicative f => (k -> a -> f b) -> MinPQueue k a -> f (MinPQueue k b)
+traverseWithKeyU' _ Empty = pure Empty
+traverseWithKeyU' f (MinPQ n k a ts) = liftA2 (\ !b !q' -> MinPQ n k b q') (f k a) (traverseForest' f (const (pure Zero)) ts)
```
konsumlamm, line 738: Forcing `q'` shouldn't make a difference, since the field is strict anyway.

src/Data/PQueue/Prio/Internals.hs
```diff
@@ -721,6 +749,18 @@ traverseForest f fCh ts0 = case ts0 of
...
+  Identity (BinomForest rk k b) #-}
+traverseForest' :: Applicative f => (k -> a -> f b) -> (rk k a -> f (rk k b)) -> BinomForest rk k a -> f (BinomForest rk k b)
+traverseForest' f fCh ts0 = case ts0 of
+  Nil       -> pure Nil
+  Skip ts'  -> (Skip $!) <$> traverseForest f fCh' ts'
+  Cons (BinomTree k a ts) tss
+    -> liftA3 (\ !p !q -> Cons (BinomTree k p q)) (f k a) (fCh ts) (traverseForest' f fCh' tss)
+  where
+    fCh' (Succ (BinomTree k a ts) tss)
+      = liftA3 (\ !p !q -> Succ (BinomTree k p q)) (f k a) (fCh ts) (fCh tss)
```
konsumlamm, line 762: Same here.

konsumlamm:
Why no `map'`, `traverse'`, `traverseU'`?

Also, do we want strict `adjustMin` etc.? It would be a lot of new functions for something that can easily be done by hand.

src/Data/PQueue/Prio/Internals.hs
treeowl: Oopsy.

src/Data/PQueue/Prio/Internals.hs
treeowl: Good point. I admit I got lazy and didn't check the constructor.

src/Data/PQueue/Prio/Internals.hs
treeowl: Uh oh. This looks worse. It's unnecessarily banged there, but misses necessary strictness in the hidden third argument.

treeowl (commented):
I don't know how many more functions to add. Do we want to do a whole `Strict` module like `containers`?

src/Data/PQueue/Prio/Internals.hs
treeowl: Fixed.

src/Data/PQueue/Prio/Internals.hs
treeowl: OK, after going through and removing the unneeded bangs, I found that it was a chore to be sure I was banging exactly the right things. In this limited context, it's totally harmless to just bang everything—GHC won't have any trouble with the extras. So I just put them back in. There are loads of places we have to be careful not to force things prematurely; this ain't one.

src/Data/PQueue/Prio/Internals.hs
treeowl: Fixed.

treeowl:
> Why no `map'`, `traverse'`, `traverseU'`?
> 
> Also, do we want strict `adjustMin` etc.? It would be a lot of new functions for something that can easily be done by hand.

For adjusting the minimum, we can't do better than

```haskell
adjustMinWithKey' f = getSolo . adjustMinWithKey ((Solo $!) . f)
```

Doing it "manually" with the constructors produces exactly the same Core, unfolding, inlining guidance, etc. So arguably we don't *need* `adjustMin'` and such, but I wouldn't be opposed to having them for user friendliness.

treeowl:
> Also, do we want strict `adjustMin` etc.? It would be a lot of new functions for something that can easily be done by hand.

I just realized I don't understand your last sentence. Implementing `adjustMin'` using `Solo` is easy for someone with enough experience, but a lot of people don't find that sort of thing obvious. I don't think it's possible to implement the strict mapping and traversal functions optimally without access to the internals. The only remaining things I can think of are indeed pretty trivial: strict insertion and strict `fromList`.

konsumlamm:
> I don't know how many more functions to add. Do we want to do a whole `Strict` module like `containers`?

> One option would be `Strict` modules that _only_ export the strict functions, and not the rest of the API. That might be less confusing than the way `containers` does it....

That might be a good idea, given the amount of strict functions. I agree that such a module should only export the strict functions. I'm also used to the strict version just having a `'` suffix though, idk how others would feel about that. Perhaps we should ask for feedback on that (e.g. on the Haskell Discourse or Reddit).

> I just realized I don't understand your last sentence. Implementing `adjustMin'` using `Solo` is easy for someone with enough experience, but a lot of people don't find that sort of thing obvious. I don't think it's possible to implement the strict mapping and traversal functions optimally without access to the internals.

Yeah, we definitely shouldn't expect everyone to know that you can use `Solo` here. I meant using `minViewWithKey` and `insert` (or the new pattern synonyms), but that wouldn't be O(1).

> The only remaining things I can think of are indeed pretty trivial: strict insertion and strict `fromList`.

For those I'd say let the user force the elements.

treeowl:
I'll ask around and see what people think. If we go with a separate module, strict insertion wouldn't be a big deal to include. Even strict extraction could make sense in that context—I wish `containers` had done that, but it's too late to change now.

```haskell
minViewWithKey q >>= \((k, v :: Int), q') ->
  Just (... [v + 1])
```

will produce a thunk to calculate `v+1`, whereas with

```haskell
minViewWithKey' q = case minViewWithKey q of
  r@(Just ((_, !_), _)) -> r
  Nothing -> Nothing
```

this will not.

```haskell
minViewWithKey' q >>= \((k, v :: Int), q') ->
  Just (... [v + 1])
```

konsumlamm:
I opened a discussion on Discourse: https://discourse.haskell.org/t/organization-of-strict-functions/6354.

PR #103, opened 2023-04-03
Strictify maps

### LINKED ISSUE #100: Make mapU/mapMonotonic and Prio.mapKeysMonotonic spine-strict
I closed #95 because `seqSpine` is potentially useful to force the spine after `mapU`/`mapMonotonic` or `Prio.mapKeysMonotonic`. But thinking about it today, I realized that those operations should almost certainly be spine-strict anyway. There's almost nothing useful you can do to a queue without potentially forcing its spine. So I'd like to make those operations force the spine, and then bring #95 back with better documentation.

treeowl:
* Make mapping force spines.
* Deprecate `seqSpine`. It shouldn't be needed any more.

Closes #100

### DIFF
--- a/src/BinomialQueue/Internals.hs
+++ b/src/BinomialQueue/Internals.hs
@@ -536,8 +536,8 @@ instance Functor rk => Functor (BinomTree rk) where
 
 instance Functor rk => Functor (BinomForest rk) where
   fmap _ Nil = Nil
-  fmap f (Skip ts) = Skip (fmap f ts)
-  fmap f (Cons t ts) = Cons (fmap f t) (fmap f ts)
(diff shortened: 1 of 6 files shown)


konsumlamm:
Are you sure those are all mapping functions that need modification? What about `Data.PQueue.Prio.Internals.mapForest` (although it seems that one isn't even used anywhere) and `Data.PQueue.Min.keysF`?

treeowl:
> Are you sure those are all mapping functions that need modification? What about `Data.PQueue.Prio.Internals.mapForest` (although it seems that one isn't even used anywhere) and `Data.PQueue.Min.keysF`?

Good catches. I just fixed those.

konsumlamm (approved):

PR #106, opened 2023-04-18
Eq and Ord instances for Prio queues

treeowl:
* Make the `Eq` and `Ord` instances for `Prio` queues compare the queues in fully sorted form—that is, as key-value maps. This seems to be the only way to make these instances make any real kind of sense.

* Document the "nondeterministic" nature of `Prio` queues.


treeowl:
@konsumlamm Any comments?

src/Data/PQueue/Prio/Internals.hs
```diff
@@ -343,9 +325,11 @@ minViewWithKey (MinPQ n k a ts) = Just ((k, a), extractHeap n ts)
...
 mapWithKey f = runIdentity . traverseWithKeyU (Identity .: f)
 
--- | \(O(n)\). @'mapKeysMonotonic' f q == 'mapKeys' f q@, but only works when @f@ is strictly
--- monotonic. /The precondition is not checked./ This function has better performance than
--- 'mapKeys'.
+-- | \(O(n)\). @'mapKeysMonotonic' f q == 'Data.PQueue.Prio.Min.mapKeys' f q@,
+-- but only works when @f@ is strictly monotonic.
+-- /The precondition is not checked./
+-- This function has better performance than 'Data.PQueue.Prio.Min.mapKeys'.
+
```
konsumlamm, line 332:
```suggestion
-- | \(O(n)\). @'mapKeysMonotonic' f q == 'Data.PQueue.Prio.Min.mapKeys' f q@,
-- but only works when @f@ is strictly monotonic.
-- /The precondition is not checked./
-- This function has better performance than 'Data.PQueue.Prio.Min.mapKeys'.
```

konsumlamm (commented):
Sorry, I don't have much time lately, I'll try to keep up with your PRs.

I'm a bit skeptical about requiring `Ord a` for `Eq (MinPQueue k a)`. That would be a breaking change and many value types I can imagine don't necessarily have an `Ord` instance (although I can't think of many use cases for an `Eq` instance anyway...). I'd be fine with just documenting that the `Eq` instance is non-deterministic and recommending ``(==) `on` toAscList``. However, then we also can't change the `Ord` instance, since they should be compatible.

Btw, please try to link related issues, like #78. Is this supposed to close #78?

treeowl:
I really want the `Eq` instance to actually mean something. I think the original idea, when it was (mistakenly) thought that `insertBehind` could be implemented efficiently, was that `Eq` would be based on the order elements come out of the queue, and `merge` would be nondeterministic. But since `insertBehind` was a failure, I think we really want to make `Eq` equality of queues-as-maps, with merges considered deterministic. It's fine to require `Ord`. Is the instance really useful? Maybe not very, but the current one is nonsense. Let's do this please.

treeowl:
Yes this should close #78.

src/Data/PQueue/Min.hs
```diff
@@ -128,7 +128,7 @@ pattern Empty = Internals.Empty
 infixr 5 :<
 
 -- | A bidirectional pattern synonym for working with the minimum view of a
--- 'MinPQueue'.  Using @:<@ to construct a queue performs an insertion in
+-- 'Prio.MinPQueue'.  Using @:<@ to construct a queue performs an insertion in
```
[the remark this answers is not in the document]
treeowl: Yes, I'll fix that.

src/Data/PQueue/Prio/Internals.hs
treeowl: What am I missing?

konsumlamm: The change is to remove the blank line, unfortunately the diff doesn't show this.

PR #107, opened 2023-04-19
Suspend less on extraction

treeowl:


src/BinomialQueue/Internals.hs
```diff
@@ -142,7 +142,9 @@ instance Ord a => Ord (MinQueue a) where
 -- The Skip constructor must be lazy to obtain the desired amortized bounds.
 -- The forest field of the Cons constructor /could/ be made strict, but that
 -- would be worse for heavily persistent use and not obviously better
--- otherwise.
+-- otherwise. In particular, if Cons is strict, then inserting into a forest
+-- that begins with a Skip will force the next node, which otherwise wouldn't
+-- be involved in the operation and which may or may not be in the CPU cache.
```
konsumlamm, line 147: What does "if Cons is strict" mean?

src/BinomialQueue/Internals.hs
```diff
@@ -331,9 +333,21 @@ incrExtract :: Extract (Succ rk) a -> Extract rk a
 incrExtract (Extract minKey (Succ kChild kChildren) ts)
   = Extract minKey kChildren (Cons kChild ts)
 
+-- Note: We used to apply Skip lazily here, for fear that the incr application
+-- might cause a cascade of thunks to be forced. However, that is not actually
+-- possible. If ts starts with a Skip, then we simply replace it without
+-- forcing its child. The only way the forest in an 'Extract' can start with a
+-- Cons is if it was produced by 'incrExtract'. But in that case, its next node
+-- came from an 'Extract', and is therefore already in WHNF. So if the 'incr'
```
konsumlamm, line 341: How can we see that the next node is already in WHNF? Is the spine of the `BinomForest` in `Extract` always forced?

konsumlamm (commented):
The benchmarks suggest that this change leads to a ~10% performance improvement.

src/BinomialQueue/Internals.hs
treeowl: `incrEctract` gets a node out of `Extract` (which must be in WHNF because `Extract` is a strict constructor) and applies `Cons` to it.

src/BinomialQueue/Internals.hs
treeowl: I mean "If Cons were strict"; sorry.

src/BinomialQueue/Internals.hs
konsumlamm: And `incrExtract` is the only function that might put a `Cons` in an `Extract`?

treeowl: That's what it looks like to me. I certainly wouldn't mind if you double checked. However, as long as you're running benchmarks, could you see what happens if we embrace the cascade and switch `incr` to `incr'` here?

konsumlamm: It looks like that to me as well. Let's add that to the comment.

Using `incr'` has more mixed results (results vary a bit between runs), but overall seems to further improve things (10-20% less than just using `$!`). Wouldn't that increase the runtime complexity of `incrExtract'` from O(1) to O(n) though?

treeowl: No, it would increase it from $O(1)$ to $O(\log n)$. But the sum total of all the `incrExtract'`s would remain $O(\log n)$, because of the way the carries work out. I'm pretty sure the worst case scenario is when the `BinomHeap` is one more than a power of two and we delete from the big tree. In that case, we'll spread the big tree all the way across the spine, then propagate carries all the way back. In other situations, we'll hit one or more `Cons`es along the way, which will lead to propagating in multiple segments, more likely to be in L1.

treeowl: Should I just switch to `incr'`? I'm fine with that.

konsumlamm: If you think that wouldn't change `incrExtract'`s complexity, that's fine by me.

treeowl: I'm confident it's fine for big $O$. Each cascade leaves a `Skip` at its head which will interrupt the next (shallower) cascade. Okasaki explains this particular thing somewhere or other in his book.

src/BinomialQueue/Internals.hs
```diff
@@ -331,9 +332,17 @@ incrExtract :: Extract (Succ rk) a -> Extract rk a
 incrExtract (Extract minKey (Succ kChild kChildren) ts)
   = Extract minKey kChildren (Cons kChild ts)
 
+-- Note: We used to apply Skip lazily here, and to use the lazy incr, for fear
+-- that the potential cascade of carries would be more expensive than leaving
+-- those carries suspended and letting subsequent operations force them.
+-- However, konsumlamm's benchmarks indicated that doing these strictly was
```
konsumlamm, line 338:
```suggestion
-- However, our benchmarks indicated that doing these strictly was
```

treeowl: If that's what you prefer. Done!

konsumlamm (approved):

Issue #94, opened 2023-03-17
Mark functions as inline/inlinable


konsumlamm:
Currently it seems that many functions take `Ord` dictionaries, even with optimizations. We should mark some functions as inline/inlinable, so that the dictionaries can be inlined.

PR #111, opened 2023-04-28
Fix mapMaybe and mapEither

### LINKED ISSUE #110: mapMaybe and mapEither are completely wrong
It looks like someone took (very clever; rather tricky to understand) code for `filter` and `partition` (which would also work for the plain `mapMaybe` in the `Prio` case) and generalized it to `mapMaybe`, `mapEither`, `mapMaybeWithKey`, and `mapEitherWithKey`. Unfortunately, the generalization is totally bogus; the order of keys in the result is not, in general, related to the order of keys in the argument.

treeowl:
* Fix incorrect behavior of `mapMaybe` and `mapEither` for `MinQueue`.

To do:

- [x] Add tests.
- [x] Port all these changes to `MinPQueue`.

Fixes #110.

### DIFF
--- a/src/BinomialQueue/Internals.hs
+++ b/src/BinomialQueue/Internals.hs
@@ -19,6 +19,7 @@ module BinomialQueue.Internals (
   minView,
   singleton,
   insert,
+  insertEager,
   union,
   unionPlusOne,
   mapMaybe,
@@ -209,6 +210,14 @@ singleton x = MinQueue (Cons (tip x) Nil)
 insert :: Ord a => a -> MinQueue a -> MinQueue a
 insert x (MinQueue ts) = MinQueue (incr (tip x) ts)
 
+-- | \(O(\log n)\), but a fast \(O(1)\) average when inserting repeatedly in
+-- an empty queue or at least around \(O(\log n)\) times into a nonempty one.
+-- Insert an element into the priority queue. This is good for 'fromList'-like
+-- operations.
+insertEager :: Ord a => a -> MinQueue a -> MinQueue a
+insertEager x (MinQueue ts) = MinQueue (incr' (tip x) ts)
+{-# INLINE insertEager #-}
+
 -- | Amortized \(O(\log \min(n,m))\), worst-case \(O(\log \max(n,m))\). Take the union of two priority queues.
 union :: Ord a => MinQueue a -> MinQueue a -> MinQueue a
 union (MinQueue f1) (MinQueue f2) = MinQueue (merge f1 f2)
@@ -219,11 +228,24 @@ unions = foldl' union empty
 
 -- | \(O(n)\). Map elements and collect the 'Just' results.
 mapMaybe :: Ord b => (a -> Maybe b) -> MinQueue a -> MinQueue b
-mapMaybe f (MinQueue ts) = mapMaybeQueue f (const empty) empty ts
+mapMaybe f = flip foldlU' empty $ \q a ->
+  case f a of
+    Nothing -> q
+    Just b -> insertEager b q
+-- This seems to be needed for specialization.
+{-# INLINABLE mapMaybe #-}
 
 -- | \(O(n)\). Map elements and separate the 'Left' and 'Right' results.
 mapEither :: (Ord b, Ord c) => (a -> Either b c) -> MinQueue a -> (MinQueue b, MinQueue c)
-mapEither f (MinQueue ts) = mapEitherQueue f (const (empty, empty)) (empty, empty) ts
+mapEither f = fromPartition .
+  foldlU'
+    (\(Partition ls rs) a ->
+        case f a of
+          Left b -> Partition (insertEager b ls) rs
+          Right b -> Partition ls (insertEager b rs))
+    (Partition empty empty)
+-- This seems to be needed for specialization.
+{-# INLINABLE mapEither #-}
 
 -- | \(O(n)\). Assumes that the function it is given is monotonic, and applies this function to every element of the priority queue,
 -- as in 'fmap'. If it is not, the result is undefined.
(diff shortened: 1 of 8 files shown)


treeowl:
To do:

- [ ] Add tests.
- [ ] Port all these changes to `MinPQueue`.
- [ ] Port the new `Monotonic` functions to `MaxQueue` and `MaxPQueue`, with `Antitone` names.

treeowl:
Another thing: we should verify that the monotonic versions are actually faster than the simple ones. That's not even slightly obvious to me. They're extremely hard to understand. At first, I thought they were structured to avoid unnecessary comparisons, but looking at them more closely, that doesn't seem to be the case.

treeowl:
Ah, but maybe it could be made so? I have to think this over....

treeowl:
I fixed the monotonic code to take what advantage it can of the ordering. However, it was still significantly (although modestly) slower than just using plain old `mapMaybe` and `mapEither`. So let's not add these monotonic specials, and just do the simplest possible thing. Less code, less error prone, and faster sounds like a nice trifecta to me.

treeowl:
@konsumlamm, would you please try to review this soon?

src/BinomialQueue/Internals.hs
```diff
@@ -219,11 +227,17 @@ unions = foldl' union empty
...
-mapMaybe f (MinQueue ts) = mapMaybeQueue f (const empty) empty ts
+mapMaybe f = flip foldlU' empty $ \q a ->
+  case f a of
+    Nothing -> q
+    Just b -> insertEager b q
 
 -- | \(O(n)\). Map elements and separate the 'Left' and 'Right' results.
 mapEither :: (Ord b, Ord c) => (a -> Either b c) -> MinQueue a -> (MinQueue b, MinQueue c)
-mapEither f (MinQueue ts) = mapEitherQueue f (const (empty, empty)) (empty, empty) ts
+mapEither f = (fromPartition .) . flip foldlU' (Partition empty empty) $ \(Partition ls rs) a ->
```
konsumlamm, line 237: Can we please avoid `(f .) . g`? I find that hard to understand.

src/BinomialQueue/Internals.hs
```diff
@@ -666,6 +661,7 @@ foldlU f z (MinQueue ts) = foldl_ f z ts
 -- @since 1.4.2
 foldlU' :: (b -> a -> b) -> b -> MinQueue a -> b
 foldlU' f z (MinQueue ts) = foldl'_ f z ts
+{-# INLINE foldlU' #-}
```
konsumlamm, line 664: Did you check that this improves things?

src/BinomialQueue/Internals.hs
```diff
@@ -209,6 +210,13 @@ singleton x = MinQueue (Cons (tip x) Nil)
 insert :: Ord a => a -> MinQueue a -> MinQueue a
 insert x (MinQueue ts) = MinQueue (incr (tip x) ts)
 
+-- | \(O(\log n)\), but a fast \(O(1)\) average when inserting repeatedly in
+-- an empty queue or at least around \(O(\log n)\) times into a nonempty one.
+-- Insert an element into the priority queue. This is good for 'fromList'-like
+-- operations.
```
konsumlamm, line 216: Should we also use it in `fromList`-like functions then? Or do you want to do that in a future PR?

src/BinomialQueue/Internals.hs
treeowl: That was actually a mistake, and won't change anything at all. I'll remove it.

src/BinomialQueue/Internals.hs
treeowl: We currently tend to inline it, but yes, I'd like to hunt those down.

src/BinomialQueue/Internals.hs
treeowl: Yeah, I suppose I can expand the whole thing out.

treeowl: Or would you be happy with `(fromPartition .) $ flip ...`?

konsumlamm: I think it would be better to just expand it to a lambda, that would be easier to understand.

src/BinomialQueue/Internals.hs
treeowl: Done.

treeowl:
I believe I've dealt with all your concerns.

src/BinomialQueue/Internals.hs
```diff
@@ -219,11 +228,24 @@ unions = foldl' union empty
 
 -- | \(O(n)\). Map elements and collect the 'Just' results.
 mapMaybe :: Ord b => (a -> Maybe b) -> MinQueue a -> MinQueue b
-mapMaybe f (MinQueue ts) = mapMaybeQueue f (const empty) empty ts
+mapMaybe f = flip foldlU' empty $ \q a ->
+  case f a of
+    Nothing -> q
+    Just b -> insertEager b q
+-- This seems to be needed for specialization.
```
konsumlamm, line 235: Also see #94, most functions don't seem to be specialized.

src/Data/PQueue/Prio/Min.hs
```diff
@@ -142,13 +142,6 @@ import qualified Data.PQueue.Prio.Internals as Internals
 
 import Prelude hiding (map, filter, break, span, takeWhile, dropWhile, splitAt, take, drop, (!!), null)
 
-#ifdef __GLASGOW_HASKELL__
-import GHC.Exts (build)
-#else
-build :: ((a -> [a] -> [a]) -> [a] -> [a]) -> [a]
-build f = f (:) []
-#endif
-
```
konsumlamm, line 151: Is this not used anymore or is it just to remove non-GHC stuff?

treeowl: It's not used in this module anymore. We still do it elsewhere, and this PR is not the place to deal with that.

src/BinomialQueue/Internals.hs
treeowl: Yeah, we need to deal with that. One step at a time ....

konsumlamm (approved):

PR #118, opened 2023-05-04
Make Prio queues more compact

### LINKED ISSUE #115: More compact Prio queues
For plain queues, we can trim wasted space by changing

```haskell
data Plain.Zero a = Plain.Zero
```

to

```haskell
newtype Plain.Zero a = Plain.Zero a
```

and recover the extra flexibility we need with a wrapper in front. This approach doesn't really work for `Prio` queues, because `data Zero k a = Zero !k a` needs three words for the `Zero` constructor.

In the shower just now, I realized there's probably a way to do it which is even simpler than what we can do for plain queues: store the value associated with each key as its rightmost child.

```haskell
data Succ rk k a = Succ {-# UNPACK #-} !(BinomTree k a) !(rk k a)
newtype Zero k a = Zero a -- Stores a value
data BinomTree rk k a = BinomTree !k (rk k a) -- No value field; the second field must be lazy
data BinomForest rk k a
  = Nil
  | Skip (BinomForest (Succ rk) k a)
  | Cons {-# UNPACK #-} !(BinomTree rk k a) (BinomForest (Succ rk) k a)

tip :: k -> a -> BinomTree Zero a
tip k a = BinomTree k (Zero a)

incr :: BinomTree rk a -> BinomForest rk a -> BinomForest rk a
incr t Nil = Cons t Nil
incr t (Skip f) = Cons t f
incr t1@(BinomTree k1 ts1) (Cons t2@(BinomTree k2 ts2) !f)
  | k1 <= k2
  = Skip $ incr (BinomTree k1 (Succ t2 ts1)) f
  | otherwise
  = Skip $ incr (BinomTree k2 (Succ t1 ts2)) f
```

On extraction, we wait for the value associated with the minimal key to be revealed at the top; no problem.

Where this technique might run into some minor difficulties is with unordered operations (including `mapWithKey`). I haven't worked through how those need to go yet, but they'll certainly be more complex than what we do now. Still, getting more compact queues without compromising complexity or performance of the key operations seems likely to be an overall win.

treeowl:
Store the value associated with each key as its rightmost child,
which saves one word per element.

As a result, the binomial trees must become lazy, which should be
good for maps and lazy traversals. The down side is that we will
need tag checks to know that we have realized `Succ` constructors.

Benchmarks indicate this improves performance.

Closes #115.

### DIFF
--- a/src/Data/PQueue/Min.hs
+++ b/src/Data/PQueue/Min.hs
@@ -263,6 +263,6 @@ keysF :: (pRk k a -> rk k) -> Prio.BinomForest pRk k a -> BinomForest rk k
 keysF f ts0 = case ts0 of
   Prio.Nil       -> Nil
   Prio.Skip ts'  -> Skip $! keysF f' ts'
-  Prio.Cons (Prio.BinomTree k _ ts) ts'
+  Prio.Cons (Prio.BinomTree k ts) ts'
     -> Cons (BinomTree k (f ts)) $! keysF f' ts'
-  where  f' (Prio.Succ (Prio.BinomTree k _ ts) tss) = Succ (BinomTree k (f ts)) (f tss)
+  where  f' (Prio.Succ (Prio.BinomTree k ts) tss) = Succ (BinomTree k (f ts)) (f tss)
--- a/src/Data/PQueue/Prio/Internals.hs
+++ b/src/Data/PQueue/Prio/Internals.hs
@@ -312,32 +312,42 @@ minViewWithKey (MinPQ n k a ts) = Just ((k, a), extractHeap n ts)
 mapWithKey :: (k -> a -> b) -> MinPQueue k a -> MinPQueue k b
 mapWithKey f = runIdentity . traverseWithKeyU (Identity .: f)
 
--- | \(O(n)\). @'mapKeysMonotonic' f q == 'mapKeys' f q@, but only works when @f@ is strictly
--- monotonic. /The precondition is not checked./ This function has better performance than
--- 'mapKeys'.
+-- | \(O(n)\). @'mapKeysMonotonic' f q == 'mapKeys' f q@, but only works when
+-- @f@ is (weakly) monotonic. /The precondition is not checked./ This function
+-- has better performance than 'mapKeys'.
 --
 -- Note: if the given function returns bottom for any of the keys in the queue, then the
 -- portion of the queue which is bottom is /unspecified/.
 mapKeysMonotonic :: (k -> k') -> MinPQueue k a -> MinPQueue k' a
 mapKeysMonotonic _ Empty = Empty
-mapKeysMonotonic f (MinPQ n k a ts) = MinPQ n (f k) a (mapKeysMonoHeap f ts)
+mapKeysMonotonic f (MinPQ n k a ts) = MinPQ n (f k) a $! mapKeysMonoHeap f ts
 
 mapKeysMonoHeap :: forall k k' a. (k -> k') -> BinomHeap k a -> BinomHeap k' a
 mapKeysMonoHeap f = mapKeysMonoForest Zeroy
   where
     mapKeysMonoForest :: Ranky rk -> BinomForest rk k a -> BinomForest rk k' a
     mapKeysMonoForest !_rky Nil = Nil
-    mapKeysMonoForest !rky (Skip rest) = Skip $ mapKeysMonoForest (Succy rky) rest
-    mapKeysMonoForest !rky (Cons t rest) = Cons (mapKeysMonoTree rky t) $ mapKeysMonoForest (Succy rky) rest
+    mapKeysMonoForest !rky (Skip rest) = Skip $! mapKeysMonoForest (Succy rky) rest
+    mapKeysMonoForest !rky (Cons t rest) = Cons (mapKeysMonoTree rky t) $! mapKeysMonoForest (Succy rky) rest
 
     {-# INLINE mapKeysMonoTree #-}
     mapKeysMonoTree :: Ranky rk -> BinomTree rk k a -> BinomTree rk k' a
-    mapKeysMonoTree !rky (BinomTree k ts) = BinomTree (f k) (mapKeysMonoTrees rky ts)
+    mapKeysMonoTree Zeroy (BinomTree k (Zero a)) =
+      -- We've reached a value, which we must not force.
+      BinomTree (f k) (Zero a)
+      -- We're not at a value; we force the result.
+    mapKeysMonoTree rky (BinomTree k ts) = BinomTree (f k) $! mapKeysMonoTrees rky ts
 
     mapKeysMonoTrees :: Ranky rk -> rk k a -> rk k' a
-    mapKeysMonoTrees Zeroy (Zero a) = Zero a
+    mapKeysMonoTrees Zeroy (Zero a) =
+      -- Don't force the value!
+      Zero a
+    mapKeysMonoTrees (Succy Zeroy) (Succ t (Zero a)) =
+      -- Don't force the value!
+      Succ (mapKeysMonoTree Zeroy t) (Zero a)
     mapKeysMonoTrees (Succy rky) (Succ t ts) =
-      Succ (mapKeysMonoTree rky t) (mapKeysMonoTrees rky ts)
+      -- Whew, no values; force the trees.
+      Succ (mapKeysMonoTree rky t) $! mapKeysMonoTrees rky ts
 
 -- | \(O(n)\). Map values and collect the 'Just' results.
 mapMaybeWithKey :: Ord k => (k -> a -> Maybe b) -> MinPQueue k a -> MinPQueue k b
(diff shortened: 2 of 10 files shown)


treeowl:
@konsumlamm This is obviously not quite ready to actually merge, but before I make it so, I was hoping you'd give it a glance and see if it's something you'd be okay with if indeed it improves performance.

treeowl:
Note: along with reducing the total queue size by $n$ words, this reduces the number of words allocated on insertion by $2$ (amortized) and the number of words allocated on deletion by $O(\log n)$, thanks to the slightly more compact `Cons` nodes.

src/Data/PQueue/Min.hs
```diff
@@ -263,6 +263,6 @@ keysF :: (pRk k a -> rk k) -> Prio.BinomForest pRk k a -> BinomForest rk k
 keysF f ts0 = case ts0 of
   Prio.Nil       -> Nil
   Prio.Skip ts'  -> Skip $! keysF f' ts'
-  Prio.Cons (Prio.BinomTree k _ ts) ts'
+  Prio.Cons (Prio.BinomTree k ts) ts'
     -> Cons (BinomTree k (f ts)) $! keysF f' ts'
-  where  f' (Prio.Succ (Prio.BinomTree k _ ts) tss) = Succ (BinomTree k (f ts)) (f tss)
+  where  f' (Prio.Succ (Prio.BinomTree k ts) tss) = Succ (BinomTree k (f ts)) (f tss)
```
treeowl, line 268: This function needs to be modified to drop values eagerly. Unfortunately, that requires writing a rank 0 wrapper around a general worker.

treeowl: Never mind. That's wrong. It does need to be modified, but the modification is trivial.

treeowl: Wrong again. I got the whole thing backwards; no modification is needed; it's already exactly as it needs to be.

treeowl:
@konsumlamm I believe this is ready to go.

treeowl:
I guess there's still the question of whether to let maps be lazy (better for performance in general) or whether to make them stricter (maintaining theoretical worst-case bounds). I'd lean toward the former in this instance, but I hate having to decide.

treeowl:
@konsumlamm, this still awaits your comments.

treeowl:
@konsumlamm It's been two weeks now.

src/Data/PQueue/Prio/Internals.hs
```diff
@@ -167,55 +169,21 @@ data BinomForest rk k a =
...
-  foldlWithKey'_ _f z Nil = z
-  foldlWithKey'_ f !z (Skip ts) = foldlWithKey'_ f z ts
-  foldlWithKey'_ f !z (Cons t ts) = foldlWithKey'_ f ft ts
-    where
-      !ft = foldlWithKey'_ f z t
-
-instance IFoldMap t => IFoldMap (BinomForest t) where
-  foldMapWithKey_ _f Nil = mempty
-  foldMapWithKey_ f (Skip ts) = foldMapWithKey_ f ts
-  foldMapWithKey_ f (Cons t ts) = foldMapWithKey_ f t `mappend` foldMapWithKey_ f ts
```
konsumlamm, line 206: Does this mean we can get rid of those classes?

src/Data/PQueue/Prio/Internals.hs
```diff
@@ -707,64 +700,43 @@ mapMWithKey f = go empty
           let !acc' = insertMax' k b acc
           go acc' q'
 
+-- | Natural numbers revealing whether something is 'Zero' or 'Succ'.
+type Ranky = Nattish Zero Succ
```
konsumlamm, line 704: Is `Nattish` only used for this? If so, wouldn't it be more clear to only define `Ranky`?

konsumlamm (commented):
Sorry for taking so long to review, this is not a simple PR.

I find the `Nattish` approach a bit confusing, but the numbers speak for themselves. I'm actually surprised that the benchmarks improve so much though, all that really changed is `Zero`, isn't it? I'd expect `data Zero k a = Zero` to be a single global value, so why does this make such a big difference?

Did you benchmark the folds? They look like they might get slower with this change.

treeowl:
I don't remember the numbers I got from benchmarks. What did they look like to you? The space savings aren't in the `Zero` value itself; they're in reclaiming the space we previously used to point to it (all over the place) and using that space to store attached values instead.

src/Data/PQueue/Prio/Internals.hs
treeowl: I would like to consider using it for the plain queues too.

src/Data/PQueue/Prio/Internals.hs
treeowl: Uhhh.... Yes, I imagine so.

konsumlamm:
> I don't remember the numbers I got from benchmarks. What did they look like to you?

The benchmarks take 10-20% less time for me.

> The space savings aren't in the `Zero` value itself; they're in reclaiming the space we previously used to point to it (all over the place) and using that space to store attached values instead.

I see, so we save one word per tree. This is $\log n$ in total though, not $n$, as you said, right?

konsumlamm:
Ok, the folds seem to improve as well.

treeowl:
> I see, so we save one word per tree. This is log⁡n in total though, not n, as you said, right?

Incorrect. You can count out the `Zero`s in a tree of a given size to get a better understanding. Or you can just notice that we have enough room to fit all $n$ of the values where we used to have nullary `Zero`s.

konsumlamm:
> > I see, so we save one word per tree. This is log⁡n in total though, not n, as you said, right?
> 
> Incorrect. You can count out the `Zero`s in a tree of a given size to get a better understanding. Or you can just notice that we have enough room to fit all n of the values where we used to have nullary `Zero`s.

Ah right, the `Zero`s are the leaves, not the roots, that's where my mistake was.

konsumlamm:
> I guess there's still the question of whether to let maps be lazy (better for performance in general) or whether to make them stricter (maintaining theoretical worst-case bounds). I'd lean toward the former in this instance, but I hate having to decide.

You mean lazy in the spine? What is the current situation? Does this PR change that?

treeowl:
What *were* off were my allocation calculations, which I believe I overstated. I'll have to recalculate those if anyone cares.

treeowl:
Lazy all through. If we want to be strict instead, I can do that, but the code will be more complicated and for the most part I expect slower.

konsumlamm:
So this basically reverts #103? Did you change your mind on #100? I'd be fine with lazy maps, if you say that's more efficient.

treeowl:
Well ... for now it keeps the plain (non-`Prio`) ones the same, as the strict constructors are better there. It more than reverts that for `Prio` queues, since it extends laziness into the trees (not just the spine). It all falls out of the representation change, which leads to different trade-offs.

konsumlamm:
I don't like having different laziness for `Prio` and non-`Prio` queues.

Can you explain why the new representation necessitates fully lazy maps? Why can't we just keep the old implementation?

I'm a bit worried that fully lazy maps might lead to situations where you build up a lot of computations which slow down a subsequent `deleteMin`, or will that not happen?

treeowl:
It doesn't necessitate fully lazy maps. Keeping them strict in the spine is easy, but not very useful for predictable performance. Keeping them entirely strict in the *structure* is quite possible too, but it's slightly trickier. That said, I just realized it should be somewhat less tricky with the `Nattish` maps than the old ones. We just need to stop the strict mapping at `Succy Zeroy`, and be lazy at `Zeroy`. It's still more code, but should be pretty readable. I'll give it a whirl and see how that goes.

konsumlamm:
> Keeping them strict in the spine is easy, but not very useful for predictable performance.

So why was that useful before, but not anymore?

treeowl:
The old representation had entirely strict trees, which held lazy values, so calculating the spine strictly calculated everything (except the values) strictly. The new representation uses the same fields for values as for lists of trees, so those fields must be lazy. To calculate everything but the values strictly, we must calculate strictly down to `Succy Zeroy`, then calculate lazily. I suspect this won't be a big deal. With the old auxiliary class approach, it would've required an extra "layer" of computational structure, but I don't think we'll need that now.

treeowl:
@konsumlamm I think maps should be back to the way they were now.

konsumlamm (approved):
Great work!

src/Data/PQueue/Prio/Internals.hs
```diff
@@ -342,14 +310,41 @@ minViewWithKey (MinPQ n k a ts) = Just ((k, a), extractHeap n ts)
...
 mapWithKey :: (k -> a -> b) -> MinPQueue k a -> MinPQueue k b
-mapWithKey f = runIdentity . traverseWithKeyU (Identity .: f)
+mapWithKey f = runIdentity . traverseWithKeyU (coerce f)
 
--- | \(O(n)\). @'mapKeysMonotonic' f q == 'mapKeys' f q@, but only works when @f@ is strictly
--- monotonic. /The precondition is not checked./ This function has better performance than
--- 'mapKeys'.
+-- | \(O(n)\). @'mapKeysMonotonic' f q == 'mapKeys' f q@, but only works when
+-- @f@ is (weakly) monotonic. /The precondition is not checked./ This function
+-- has better performance than 'mapKeys'.
```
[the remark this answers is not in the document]
treeowl: Correct. And makes the tests verify that

PR #125, opened 2023-06-02
Update GHC versions in haskell-ci.yml

tmcgilchrist:
ghc 9.6.1 -> 9.6.2 
ghc 9.4.4 -> 9.4.5

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -28,14 +28,14 @@ jobs:
     strategy:
       matrix:
         include:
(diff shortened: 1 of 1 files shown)


treeowl:
Thanks! Unfortunately, this isn't how we make these updates. We use `haskell-ci` to manage the `haskell-ci.yml` file. Specifically, the thing to do is update the `tested-with` field of `pqueue.cabal` and then run `haskell-ci regenerate`. I'll just do it this time, since I already sat through the interminable process of compiling the dependencies of `haskell-ci`.

PR #126, opened 2023-06-02
Use WordPtr instead of Word for Nattish

### LINKED ISSUE #124: Type safety issue?
The fake GADT we now use for maps and unordered traversals and folds is a little bit unsafe. I thought that we were well protected by the fact that the queue would have to reach `O(2^WORDSIZE)` elements to cause a problem, but it's actually possible to do this in a reasonable amount of time on systems with 32-bit `Word` but addressable memory considerably exceeding $2^32$ words, if there are any such around. In particular, an adversary could calculate something like `stimes (2^(2^32)) (singleton () ())`. This will take quite a long time, and something in excess of 64 gigabytes of memory, but I think it's possible in principle.

Suppose someone uses `traverseU` on this monstrosity, with a `Backwards` applicative. I suspect they will be able to actually reach the point where underflow occurs, leading to an unsafe coercion and memory fault.

----

The easiest solution is to use `Word64` instead of `Word`. I don't think it's realistic to produce a queue whose size exceeds $2^{2^{64}}$ by any means whatsoever. The downside is that this will hurt performance on 32-bit systems. I don't know how much we care.

The other obvious solution is to make sure we never underflow. The best way to do that is probably to perform a size check on `insert` and on `union`. This has the side benefit of ensuring that `size` always produces a correct value. The downside of course is that insertion becomes ever so slightly slower.

treeowl:
Use `WordPtr` instead of `Word` for `Nattish`. `Word` is actually fine, but I think `WordPtr` makes it more *obviously* fine, and also adds a bit of insurance against hypothetical future changes to GHC.

Closes #124.

### DIFF
--- a/src/Nattish.hs
+++ b/src/Nattish.hs
@@ -18,6 +18,7 @@ module Nattish
   ( Nattish (Zeroy, Succy)
   )
   where
+import Foreign.Ptr (WordPtr)
 import Unsafe.Coerce (unsafeCoerce)
 #if __GLASGOW_HASKELL__ >= 800
 import Data.Kind (Type)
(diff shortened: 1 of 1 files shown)


src/Nattish.hs
```diff
@@ -81,4 +82,30 @@ pattern Succy n <- (check -> ResSucc n)
...
+-- left (using a lazy left fold, for example), we could overflow. Disaster!
+-- Fortunately, we have WordPtr, which is equivalent to uintptr_t, and
+-- therefore big enough to hold the logarithm of the size of any queue
+-- represented in memory, even with maximal sharing.
+--
+-- Now, according to a comment on
+-- https://hackage.haskell.org/package/base-4.18.0.0/docs/src/Foreign.Ptr.html#WordPtr
+-- GHC actually guarantees that a Word is the same size as a pointer, so we
+-- actually *could* just use Word, but I think WordPtr makes it *utterly clear*
+-- that we're safe here.
```
konsumlamm, line 109: What? We use `WordPtr` because it could be bigger than `Word`, but actually they're always the same size? What's the point then? I think using `WordPtr` only makes it more confusing, since usually noone uses a pointer as a number type.

treeowl: On GHC they're the same. If someone ports to another implementation, that might not be.

treeowl: If you're dead set against this, I can go back to `Word` and explain why it's okay on GHC. That just commits us even more heavily to GHC *specifically*, rather than to just a compiler with similar power.

PR #137, opened 2024-05-22
Fix unused imports and other warnings

pgujjula:
This PR re-enables `Wunused-imports` and removes all unused imports.

We also use CPP to fix a few more warnings that only appear on some GHCs.


konsumlamm (approved):
Thanks!

PR #140, opened 2025-02-28
Optimize `mapMonotonic` by using an `Fmap` class instead of `Functor`

### LINKED ISSUE #55: Make sure we're not building unnecessarily large dictionaries

konsumlamm:
Resolves #55.

This is the same optimization as in #59. Due to the recursive nature of our instances, the `Functor` instance built a lot of unnecessary `<$`s, which are avoided by the new `Fmap` class. This is the only such remaining instance that I found.

This change improves the performance of `mapMonotonic` significantly (see benchmarks below).

<details>
<summary>Benchmark results</summary>

Benchmarking `mapMonotonic (+ 1)` on a `MinQueue Int` with n random elements.
Before:
```
All
  mapMonotonic
    1000:    OK
      10.8 μs ± 286 ns
    10000:   OK
      129  μs ± 7.5 μs
    100000:  OK
      3.79 ms ± 230 μs
    1000000: OK
      98.9 ms ± 6.8 ms
```
After:
```
All
  mapMonotonic
    1000:    OK
      6.69 μs ± 398 ns, 37% less than baseline
    10000:   OK
      76.6 μs ± 2.4 μs, 40% less than baseline
    100000:  OK
      3.13 ms ± 212 μs, 17% less than baseline
    1000000: OK
      85.0 ms ± 6.0 ms, 14% less than baseline
```
</details>

### DIFF
--- a/pqueue.cabal
+++ b/pqueue.cabal
@@ -58,8 +58,8 @@ library
     BinomialQueue.Internals
     BinomialQueue.Min
     BinomialQueue.Max
+    Data.PQueue.Internals.Classes
     Data.PQueue.Internals.Down
-    Data.PQueue.Internals.Foldable
     Data.PQueue.Prio.Max.Internals
     Nattish
   if impl(ghc) {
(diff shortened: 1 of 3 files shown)


treeowl:
Good catch!

treeowl (approved):

PR #141, opened 2025-03-06
More `coerce`

### LINKED ISSUE #101: Avoid eta expansion by composition
We have a lot of `f . unDown` getting passed to higher-order functions that can't inline. In situations where GHC doesn't know that `f` has arity at least 1, the RTS will allocate a closure through which to call `f`. That's not nice. We should borrow a trick from `profunctors`:

```haskell
infixl 8 .#
(.#) :: forall a b c q. Coercible b a => (b -> c) -> q a b -> a -> c
(.#) pbc _ = coerce pbc
{-# INLINE (.#) #-}

-- We may not need this one
infixr 9 #.
(#.) :: forall a b c q. Coercible c b => q b c -> (a -> b) -> a -> c
(#.) _ = coerce (\x -> x :: b) :: forall a b. Coercible b a => a -> b
{-# INLINE (#.) #-}
```

Then we can write `Down #. f` to coerce it nicely. The exact definitions in `profunctors` are carefully arranged to work around some weirdness in `Coercible` solving back in 7.10 and (IIRC) 8.0.

Then we can write things like

```haskell
traverseWithKey f (MaxPQ q) = MaxPQ <$> Q.traverseWithKey (f #. unDown) q
```

We'll also want to implement `mapWithKey` and `mapWithKey'` using coercions.

konsumlamm:
Closes #101.

This probably doesn't change much in the generated code, but at least it's more concise and we can be sure that the conversions get optimized away.

### DIFF
--- a/src/BinomialQueue/Max.hs
+++ b/src/BinomialQueue/Max.hs
@@ -88,7 +88,6 @@ module BinomialQueue.Max (
 import Prelude hiding (null, take, drop, takeWhile, dropWhile, splitAt, span, break, (!!), filter, map)
 
 import Data.Coerce (coerce)
-import Data.Bifunctor (bimap)
 import qualified Data.List as List
 import Data.Maybe (fromMaybe)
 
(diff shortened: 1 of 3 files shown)


konsumlamm:
@treeowl do you want to take a look or can I go ahead and merge?

